PR #10, opened 2017-11-02
XW-4180 | Unmute when related video was played by user interaction

bkoval:
@Wikia/x-wing


src/index.js
```diff
@@ -94,6 +94,7 @@ window.wikiaJWPlayer = function (elementId, options, callback) {
 
 		wikiaJWPlayerReplaceIcons(playerInstance);
 		wikiaJWPlayerEvents(playerInstance, options.autoplay.enabled, logger);
+		options.related && wikiaJWPlayerRelatedVideoSound(playerInstance);
```
latata: IMO `if` is more readable. 
And code probably will be minified to similar statement.

PR #13, opened 2017-11-03
XW-4191 | unmute property as a separate prop

JamesHawking:
https://wikia-inc.atlassian.net/browse/XW-4191


dist/index.html
```diff
@@ -30,6 +30,8 @@
 		},
 		// set to true if you want video to autostart
 		autoplay: true,
+		// set to false if you want to autoplay on mobile
```
xkxd: we should elaborate on that more, it looks weird now

src/index.html
```diff
@@ -30,6 +30,8 @@
 		},
 		// set to true if you want video to autostart
 		autoplay: true,
+		// set to false if you want to autoplay on mobile
```
latata: `set to true if you want to have video initially muted`

PR #14, opened 2017-11-03
XW-4192 | add captions toggle

xkxd:
https://wikia-inc.atlassian.net/browse/XW-4192

@Wikia/x-wing


src/settings.plugin.js
```diff
@@ -257,6 +246,104 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+	return captionObj.id === 'off';
+};
+
+wikiaJWPlayerSettingsPlugin.prototype.getUserLang = function() {
+	return window.navigator.language.slice(0, 2);
+};
+
+wikiaJWPlayerSettingsPlugin.prototype.getSuitableCaptions = function(userLang, captionTracks) {
+	return captionTracks
+		.map(function (track) {
```
hakubo: ({label}) => label

xkxd: no ES6 yet :<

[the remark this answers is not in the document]
xkxd: after doing some changes I still consider this cleaner than forEach / some. but I changed method name and added a comment about using Array#findIndex in the future

src/settings.plugin.js
```diff
@@ -257,6 +247,118 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+function createToggle(params) {
+	var toggleWrapper = document.createElement('li'),
+		toggleInput = document.createElement('input'),
+		toggleLabel = document.createElement('label');
+
+	toggleWrapper.className = 'wikia-jw-settings__toggle';
+
+	toggleInput.setAttribute('type', 'checkbox');
+	toggleInput.setAttribute('id', params.id);
+	toggleInput.classList.add('wds-toggle__input');
```
bkoval: I think we could leverage className here as well :)

xkxd: i forgot about those little one, thx

src/index.html
```diff
@@ -37,7 +37,8 @@
 			// set to true when you want to give user option to change quality of the video
 			// show quality option doesn't appear in Safari and mobile browsers even if the option is set to true
 			// these browsers have native HLS support
-			showQuality: true
+			showQuality: true,
+			showCaptionsToggle: true
```
latata: add description in a comment

src/settings.plugin.js
```diff
@@ -20,7 +20,8 @@ function wikiaJWPlayerSettingsPlugin(player, config, div) {
 }
 
 wikiaJWPlayerSettingsPlugin.prototype.isSettingsMenuOrSettingsButton = function (element) {
-	var button = this.player.getContainer().querySelector('[button=' + this.buttonID + ']');
+	var button = this.getSettingsButtonElement();
```
latata: 👍

src/settings.plugin.js
```diff
@@ -257,6 +247,118 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+	}
+};
+
+wikiaJWPlayerSettingsPlugin.prototype.areCaptionsOff = function(captionIndex) {
+	// "off" always has first position in tracks array
+	return captionIndex === 0;
+};
+
+wikiaJWPlayerSettingsPlugin.prototype.getUserLang = function() {
+	return window.navigator.language.slice(0, 2);
```
latata: Does IE11 support it? 
`https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language`

xkxd: from what i read - it should, but I've added fallback anyway

src/index.js
```diff
@@ -64,15 +64,22 @@ window.wikiaJWPlayer = function (elementId, options, callback) {
 				image: '//content.jwplatform.com/thumbs/' + videoId + '-640.jpg',
 				mute: options.mute,
 				playlist: options.videoDetails.playlist,
-				title: options.videoDetails.title
+				title: options.videoDetails.title,
+				captions: {
```
latata: Shouldn't it be override by styles property if provided?

src/settings.plugin.js
```diff
@@ -259,6 +249,125 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+		if (event.tracks.length > 1) {
+			this.currentlySelectedCaptions = this.getSuitableCaptionsIndex(
+				this.captionLangMap[this.getUserLang()],
+				event.tracks
+			);
+
+			this.getLabelElement(this.captionsToggle)
+				.addEventListener('click', clickHandler);
+
+			this.wikiaSettingsElement.classList.remove('are-captions-empty');
```
latata: How about extracting `'are-captions-empty'` to variable as it's used twice

src/settings.plugin.js
```diff
@@ -259,6 +249,125 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+			this.show();
+
+			if (this.config.captions.enabled) {
+				this.player.setCurrentCaptions(this.currentlySelectedCaptions);
+			}
+
+			if (this.config.captions.styles && Object.keys(this.config.captions.styles).length) {
```
latata: Should it be done just in player setup?

src/settings.plugin.js
```diff
@@ -259,6 +249,125 @@ wikiaJWPlayerSettingsPlugin.prototype.updateCurrentQuality = function (data) {
...
+	this.captionsToggle.classList.add('wikia-jw-settings__captions-button');
+
+	return this.captionsToggle;
+};
+
+wikiaJWPlayerSettingsPlugin.prototype.captionsClickHandler = function () {
+	var nextSelectedTrack = this.areCaptionsOff(this.player.getCurrentCaptions()) ? this.currentlySelectedCaptions : 0;
+
+	this.player.setCurrentCaptions(nextSelectedTrack);
+	this.player.trigger('captionsSelected', {
```
latata: Do we want to track it?

src/index.html
```diff
@@ -32,14 +32,24 @@
...
 			// autoplay toggle appears in settings menu which sends event `autoplayToggle` on click
 			showAutoplayToggle: true,
 			// set to true when you want to give user option to change quality of the video
 			// show quality option doesn't appear in Safari and mobile browsers even if the option is set to true
 			// these browsers have native HLS support
-			showQuality: true
+			showQuality: true,
+			// set to true when you want to show captions toggle
+			// captions toggle appears in settings menu, sends event `captionsSelected` on click
+			showCaptionsToggle: true
```
latata: It won't be just a toggle soon so I'd rename it to `showCaptions`

xkxd: note for renaming this in specific PR

src/index.js
xkxd: it's not inlined so it's overridden by css anyway, that's why I'll aim to style them via `::cue` in css.

src/settings.plugin.js
xkxd: will be fixed in captions-list PR

src/settings.plugin.js
xkxd: true, will be fixed in captions-list PR

src/settings.plugin.js
xkxd: I'll ask Stan

PR #16, opened 2017-11-08
XW-4229 | i18n

latata:
https://wikia-inc.atlassian.net/browse/XW-4229
@Wikia/x-wing


scripts/i18n-build.js
```diff
@@ -0,0 +1,19 @@
+var fs = require('fs');
```
hakubo: we use node 6 everywhere now, so you can use es6 here

src/locales/pt/main.json
```diff
@@ -0,0 +1,24 @@
...
+  "nextUp": "Next Up", 
+  "nextUpInSeconds": "Next up in xx", 
+  "pause": "Pause", 
+  "play": "Play", 
+  "playback": "Start playback", 
+  "player": "Video Player", 
+  "prev": "Previous", 
+  "replay": "Replay", 
+  "settings": "Settings", 
+  "skipmessage": "Skip ad in xx",
```
hakubo: we should document it in crowding so they don't translate xx :)

scripts/i18n-build.js
```diff
@@ -0,0 +1,19 @@
...
+languages.forEach(function (lang, index) {
+	var stat = fs.statSync(path + lang);
+	if (stat.isDirectory()) {
+		var translations = fs.readFileSync(path + lang + '/main.json').toString();
+		result += '"' + lang + '": ' + translations;
+		if (index < languages.length - 1) {
+			result += ',\n';
+		}
+	}
+});
```
xkxd: with `Array#reduce` you can avoid that `result` variable.

src/index.js
```diff
@@ -53,26 +53,37 @@ window.wikiaJWPlayer = function (elementId, options, callback) {
 		var playerInstance = jwplayer(elementId),
 			videoId = options.videoDetails.playlist[0].mediaid,
 			willAutoplay = options.autoplay,
+			lang = options.lang || 'en',
+			// IMA supports two-letter ISO 639-1 code
+			langForAds = lang.substr(0, 2),
+			i18n = wikiaJWPlayeri18n[lang],
```
xkxd: i'd rename `i18n` to `langObj`

latata: this name is consistent with mobile wiki/mercury i18n helper name :) so I'd leave it as is

scripts/i18n-build.js
latata: I think that it won't reduce number of variables used. And for me old good `forEach` is more readable.

PR #22, opened 2017-11-09
XW-4118 | Do not track video seconds played

bkoval:
@Wikia/x-wing


src/tracking.js
```diff
@@ -149,13 +149,6 @@ function wikiaJWPlayerTracking(playerInstance, willAutoplay, tracker) {
 		});
 	});
 
-	playerInstance.on('videoSecondsPlayed', function (data) {
```
latata: I'd remove `videoSecondsPlayed`events from events.js as well.

PR #23, opened 2017-11-14
webpack-based build

xkxd:
pros
* es6 + modules
* live reload
* no dev.css in repo
* no jsons concatenation
* source maps
* single index.html

cons
* gzipped size went up from 6,5KB to 7 KB  
* \<add additional ones\>


package.json
```diff
@@ -29,13 +18,20 @@
...
   "devDependencies": {
-    "autoprefixer": "7.1.6",
-    "design-system": "git+https://github.com/Wikia/design-system.git#8.11.1",
-    "http-server": "0.10.0",
-    "node-sass": "4.5.3",
-    "nodemon": "1.12.1",
-    "npm-run-all": "4.1.1",
-    "postcss-cli": "4.1.1",
-    "uglify-js": "3.1.5"
+    "autoprefixer": "^7.1.6",
```
latata: I'd keep versions without `^`

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+    ["env", {
+      "useBuiltIns": true,
+      "targets": {
+        "browsers": ["last 2 versions", "safari >= 7", "ie 10"]
```
bkoval: IE11+

package.json
```diff
@@ -1,22 +1,11 @@
 {
   "name": "jwplayer-fandom",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
bkoval: Are we sure about it? ;>

postcss.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+	plugins: [
+		require('autoprefixer')({})
```
bkoval: Would that somehow share babel config for supported browsers? If not, we should probably add it at some point :)

PR #24, opened 2017-11-15
ADEN-6229 Fix controls on VPAID

fraszczakszymon:


src/styles/overrides.scss
```diff
@@ -64,6 +64,7 @@
 	.jw-controlbar {
 		// background relies on player's id selector so it's impossible to set it for every case
 		background: linear-gradient(0, #000, transparent) !important;
+		z-index: 2;
```
latata: I'd add a comment why we set z-index

PR #25, opened 2017-11-20
XW-4271 | Change fallback for zh language

rybmat:
https://wikia-inc.atlassian.net/browse/XW-4271

Based on discussion in comments of this ticket: https://wikia-inc.atlassian.net/browse/LOC-17741 we should change zh language code to fallback to zh-hans instead of zh-tw. Also new translations are available

@Wikia/x-wing


package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "npm run build && npm-run-all --parallel watch-other watch-js watch-scss watch-i18n server",
     "server": "http-server -c-1 dist",
     "copy-html": "mkdir -p dist && cp src/*.html dist/",
-    "build": "npm run copy-html && npm run build-all-css && npm run build-js && npm run build-i18n",
+    "build": "npm run copy-html && npm run build-i18n && npm run build-all-css && npm run build-js",
```
bkoval: Why the change of ordering?

latata: Because `build-i18n` build `i18n.js` file in src directory which is then used by `build-js` script

PR #33, opened 2018-02-22
XW-4668 | Watermark with fandom-logo

bkoval:
@Wikia/x-wing


src/watermark.plugin.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * prepares watermark link with icon
+ */
+WikiaJWPlayerWatermarkPlugin.prototype.getWatermarkElement = function () {
+	var watermarkImage = wikiaJWPlayerIcons['fandomLogo'];
+
+	var watermarkElement = document.createElement('a');
+	watermarkElement.classList.add('wikia-watermark');
+	watermarkElement.innerHTML = watermarkImage;
+	watermarkElement.href = 'https://fandom.com';
```
rybmat: probably not for now, but we could parametrise this link somehow, so video team could specify where specific wordmark redirects.

src/index.js
```diff
@@ -85,6 +86,9 @@ window.wikiaJWPlayer = function (elementId, options, callback) {
 					autoplay: options.autoplay,
 					selectedCaptionsLanguage: options.selectedCaptionsLanguage,
 					i18n: i18n
+				},
+				wikiaWatermark: {
+					show: true//!!options.videoDetails.playlist[0].watermark
```
latata: I think this option should just activate watermark plugin. If plugin is active then show watermark if `watermark` custom parameter is set. If plugin is not active we shouldn't show watermark even if the custom parameter is set.

src/watermark.plugin.js
```diff
@@ -0,0 +1,60 @@
...
+	if (this.watermarkElement.parentElement) {
+		this.container.innerHTML = '';
+	}
+};
+
+/**
+ * shows back the entire plugin (adds button back)
+ */
+WikiaJWPlayerWatermarkPlugin.prototype.show = function () {
+	if (!this.watermarkElement.parentElement && this.isEnabled) {
```
latata: It looks like we will show watermark for all videos, not just for these with `watermark` custom param set.

src/watermark.plugin.js
```diff
@@ -0,0 +1,60 @@
...
+	this.container = div;
+	this.config = config;
+	this.watermarkElement = this.getWatermarkElement();
+	this.watermarkElement.addEventListener('click', function () {
+		player.trigger('watermarkClicked');
+	});
+
+	this.container.classList.add('wikia-watermark-container');
+
+	this.isEnabled = config.show;
```
latata: I think this should be the value of `watermark` custom parameter from the current video.

src/watermark.plugin.js
bkoval: Yeah, it will be easy to do so, once we have more than one watermark :)

PR #46, opened 2018-03-27
XW-4133 | Fix tracking

hakubo:
due to incorrect rounding for some videos we were not tracking certain percentages

this should ensure that we send all tracking both seconds and percentage

by rounding progress to nearest 5


src/events.js
```diff
@@ -35,47 +35,65 @@ function wikiaJWPlayerEvents(playerInstance, willAutoplay, logger) {
...
 			playerInstance.trigger(prefix + 'MidPoint');
 			state[prefix].wasMidPointTriggered = true;
 		}
 
-		if (percentPlayed >= 75 && !state[prefix].wasThirdQuartileTriggered) {
+		if (percentPlayedRounded >= 75 && !state[prefix].wasThirdQuartileTriggered) {
 			playerInstance.trigger(prefix + 'ThirdQuartile');
 			state[prefix].wasThirdQuartileTriggered = true;
 		}
+		console.log(percentPlayedRounded, state[prefix].progress.percentWatched);
```
rybmat: console log?

src/events.js
```diff
@@ -35,47 +35,64 @@ function wikiaJWPlayerEvents(playerInstance, willAutoplay, logger) {
...
+	function roundPositionProgress(position) {
+		return Math.floor(position / 5) * 5;
+	}
+
+	/**
+	 * Rounds progress in percents to nearest 5
+	 * @param {number} position 
+	 * @param {number} duration 
+	 */
+	function roundPercentProgress(position, duration) {
```
xkxd: consider 
```
round = (number) => {
    return Math.floor(number / 5) * 5;
}
```
and 
```
var positionRounded = round(data.position),
var percentPlayedRounded = round((data.position / data.duration) * 100);
```

PR #47, opened 2018-04-06
XW-4735 | Debug level for logger

bkoval:
@Wikia/x-wing


src/logger.js
```diff
@@ -38,12 +41,21 @@ function wikiaJWPlayerLogger(options) {
...
 		request.send(JSON.stringify(data));
 	}
 
+	function debug(name, description) {
+		if (logLevel <= logLevels['debug']) {
+			console.log(prefix, name, description);
+		}
+
+		if (logDebugToService) {
+			logErrorToService('debug', name, description);
```
latata: there is no such method anymore. should be `logToService `

PR #54, opened 2018-10-29
IW-130 | User-Intended plays recognition module

bkoval:
@Wikia/iwing


src/events.js
```diff
@@ -91,7 +91,6 @@ function wikiaJWPlayerEvents(playerInstance, willAutoplay, logger) {
 		});
 
 		relatedPlugin.on('play', function (data) {
-			logger.info('related plugin play');
```
bkoval: Looks like a dev leftover from the past, cleanups.

src/index.js
```diff
@@ -154,6 +155,14 @@ window.wikiaJWPlayer = function (elementId, options, callback) {
 
 		wikiaJWPlayerAllowControllOnTouchDevices(playerInstance);
 
+		wikiaJWPlayerUserIntendedPlayControl(
+			//user coming from footer must've clicked the video thumbnail
+			window.location.search.indexOf('wikia-footer-wiki-rec') > -1,
```
rybmat: not sure if we should do it now, but how about extracting list of such queryparameters to some array? I assume there will be more places we would count as UIP

src/user-intended-play-control.js
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	function onPause(data) {
+		if (data.pauseReason === 'interaction') {
+			wasPausedByUserInteraction = true;
+		}
+	}
+
+	function onPlay(data) {
```
latata: Unused `data`

src/user-intended-play-control.js
```diff
@@ -0,0 +1,104 @@
...
+	function onFullScreen() {
+		setUserIntendedPlay(true, reasonsForUserIntendedPlay.fullScreen);
+	}
+
+	function onUnmute() {
+		setUserIntendedPlay(true, reasonsForUserIntendedPlay.unmuting);
+	}
+
+	function onMute() {
+		if (isUserIntendedPlay && userIntendedPlayReason === reasonsForUserIntendedPlay.unmuting) {
```
latata: As we talked offline please get rid of `reasonsForUserIntendedPlay` and add flag which determines if the reason of user intended play was unmuting. Or even we can get rid of this `if` completely. Ask product.

src/user-intended-play-control.js
```diff
@@ -0,0 +1,104 @@
...
+		} else {
+			userIntendedReason = null;
+		}
+
+		if (typeof tracker.setCustomDimension !== 'function') {
+			return;
+		}
+
+		if (immediate) {
+			tracker.setCustomDimension(39, isUserIntended ? 'user-intended' : 'not-user-intended');
```
latata: lets extract CD number and values to consts

src/index.js
bkoval: That makes sense but we I will actually move this logic to app repository and implement it there :)

PR #56, opened 2019-01-21
IW-1231 | Unmute and autoplay FV on tab switch

mszabo-wikia:
https://wikia-inc.atlassian.net/browse/IW-1231

@Wikia/iwing


package.json
```diff
@@ -40,7 +40,7 @@
     "autoprefixer": "7.1.6",
     "design-system": "git+https://github.com/Wikia/design-system.git#8.12.0",
     "http-server": "0.10.0",
-    "node-sass": "4.5.3",
+    "node-sass": "4.11.0",
```
mszabo-wikia: build failed for me otherwise

src/tab-active.js
```diff
@@ -1,19 +1,20 @@
...
-	}
-
+function wikiaJWPlayerHandleTabNotActive(playerInstance, willAutoplay) {
 	var pausedOnRelated = false;
 
-	playerInstance.on('playerStart', function () {
-		document.addEventListener('visibilitychange', function () {
-			if (canPlayVideo()) {
+	function playVideoOnTabSwitch() {
+		if (['playing', 'paused', 'complete'].indexOf(playerInstance.getState()) === -1 || pausedOnRelated) {
```
latata: Don't we need `document.hidden` check anymore? I looks like it can start play now when someone is leaving the tab.

src/tab-active.js
```diff
@@ -1,19 +1,20 @@
...
+			if (willAutoplay) {
 				playerInstance.play();
 				pausedOnRelated = false;
 				playerInstance.trigger('playerResumedByBrowserTabSwitch');
+			} else {
+				playerInstance.setMute(false);
```
latata: Why do we need this? IMO for click-to-play videos the mute property in the options object should be set to false (https://github.com/Wikia/jwplayer-fandom/blame/master/src/index.js#L83)

mszabo-wikia: unfortunately it does not work in this situation, video starts muted when you click play

PR #63, opened 2020-06-23
IW-3321 | play next when the video is geo-restricted, yarnify repo

latata:
https://wikia-inc.atlassian.net/browse/IW-3321


src/events.js
```diff
@@ -146,4 +146,12 @@ function wikiaJWPlayerEvents(playerInstance, willAutoplay, logger) {
 			playerInstance.trigger('jwplayerClosed')
 		}
 	});
+
+	playerInstance.on('error', function (data) {
+		// Play next video if the user has no rights to play the current one
+		// This code is usually returned when the video is geo restricted
+		if (depth > 0 && data.code === 232403) {
+			playerInstance.next();
```
iraminius, line 154: So what's the possibility of getting the same error on the next video? 😄 Also, what `depth` means here?

latata: It's possible, it will play the next one then. The `depth` means how many times the next video was played. `0` means it's the initial video.

iraminius: So if there aren't any more videos then it won't play it, right?
And why aren't we skipping the initial video if it's geo-restricted?

latata: I think it might be confusing for the user that he gets the video which is not relevant to the content because the one that is relevant is geo-restricted.

PR #74, opened 2021-08-02
PIE-977: Utilize specific alternate JW Player ID for Android based traffic

gundams-and-dragons:
https://wikia-inc.atlassian.net/browse/PIE-977


gundams-and-dragons:
That's just the file that gets built and used.

PR #79, opened 2022-01-11
Fandom wirewax plugin

gundams-and-dragons:
Purpose: This is creating a plugin that loads the embedder SDK, binds the jwplayer and wirewax events, and sets up tracking.

Reference: https://devhub.wirewax.com/docs/fandom-sdk/#full-template


src/fandom-wirewax.plugin.js
```diff
@@ -0,0 +1,257 @@
...
+  });
+};
+
+function FandomWirewaxPlugin(rootId, options) {
+  this.isPlayerRegistered = false;
+  this.rootId = rootId;
+  this.options = options;
+  this.player = options.player;
+  this.autoPlay = true;
+  this.setupEmbedder = this.setupEmbedder.bind(this);
```
harageth, line 32: All of these functions are defined in the source in fandomSDKUrl for needing to be overridden?

src/fandom-wirewax.plugin.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+FandomWirewaxPlugin.prototype.setupEmbedder = function () {
+  if (!this.embedder) {
+    // Create a container
+    this.container = document.createElement("div");
+    this.container.classList.add("vjs-wirewax-container");
+    this.container.setAttribute(
+      "style",
+      "position: absolute; height: 100%; width: 100%; top: 0; pointer-events: none"
```
harageth, line 104: We are doing this inline like this because its the only item and easier than adding a .css file?

src/fandom-wirewax.plugin.js
```diff
@@ -0,0 +1,257 @@
...
+    this.startTimeUpdate();
+    this.embedder.play();
+  }
+
+  if (this.isPlayerRegistered) return;
+
+  window.dataLayer = window.dataLayer || [];
+
+  // Bind WIREWAX to JW player events
+  this.player.on("play", this.JWPlayHandler);
```
harageth, line 145: Whats the difference between JWPlayHandler and the WirewaxPlayHandler? I know one is the player but I dont understand why we need to use both

harageth (approved):
Left a few questions because I dont understand some of the frontend stuff. As far as I can tell everything makes sense to me. Code styling is good and it looks to be doing what I expect this to be doing.

harageth:
Should probably get one more review from someone else with more frontend experience

src/fandom-wirewax.plugin.js
```diff
@@ -0,0 +1,257 @@
+// fandom-jw-plugin.js
+
+console.log("FandomWirewaxPlugin loaded");
```
JoshuaRogan, line 3: Debugging?

JoshuaRogan (approved):

src/fandom-wirewax.plugin.js
```diff
@@ -0,0 +1,257 @@
...
+    console.warn(error);
+  }
+};
+
+FandomWirewaxPlugin.prototype.JWPauseHandler = function (event) {
+  this.stopTimeUpdate();
+
+  try {
+    this.embedder.pause();
+    window.dataLayer.push({ event: 'wirewax', action: 'pause', eventData: event });
```
JoshuaRogan, line 186: This is a great example of an area that would benefit from having the function we created in pathfinder. (later though)

src/fandom-wirewax.plugin.js
harageth: From the few modules I've looked at this seems like fairly standard practice

vforge: Please, no, we have enough in console already; can it be at least debug?

src/fandom-wirewax.plugin.js
gundams-and-dragons: Also the div being targeted only exists when wirewax is embedded.

src/fandom-wirewax.plugin.js
gundams-and-dragons: These are mostly for use in the embedding/handler functions.

PR #82, opened 2022-04-15
DPAD-262 | Add tracking and record events + minor improvements 

JoshuaRogan:
* Added editorconfig file that matches eslintc/prettier for better ide support
* Update nvmrc file to v17
* Updated npmrc file 
* Installed tracking library 
* Added many tracking events
* Add test-jw create-react-app for in repo testing

@gundams-and-dragons


yarn-error.log
```diff
@@ -1,38 +1,36 @@
 Arguments:
```
vforge, line 1: can we remove and gitignore this file?

JoshuaRogan: Yup, looks like there are some others too

JoshuaRogan: https://github.com/Wikia/jwplayer-fandom/pull/82/commits/c7431adf5dc15f6716ac3cf565bc276c9e9afb5d

test-jw/src/App.js
```diff
@@ -0,0 +1,11 @@
+import { VideoPlayer } from './bundle.esm'
```
gundams-and-dragons, line 1: This is loading a different bundle.esm than the one being symlinked.

JoshuaRogan: The symlink command is `ln -s dist/bundle.esm.js test-jw/ ` which maps to `./bundle.esm` don't need to add the .js extension. Am I missing something else?

p-forfa (approved):
Approving for merge to integration branch. The changes look good with the new test-jw folder, but I didn't test it. It looks similar to what was implemented in a temp branch in the video-platform repo.

It does look like a new bundle.esm.js file is put into the test-jw/src folder. Maybe it should use the one in the /dist directory instead, unless that's the bundle for the test app itself.

PR #86, opened 2022-04-21
Use npm and remove dist directory

JoshuaRogan:
* Added scripts for publishing 
* Remove dist directory 🔥 
* Changed the root source from dist to / because npm will publish _from_ the dist directory
* ~Husky doesn't play well when publishing in this format. Removed for now, the PR checks should keep us safe for now but I would like to find a fix to add it back~ fixed


package.json
```diff
@@ -1,13 +1,13 @@
...
-	"main": "dist/bundle.umd.js",
-	"module": "dist/bundle.esm.js",
-	"types": "dist/main.d.ts",
+	"main": "bundle.umd.js",
+	"module": "bundle.esm.js",
+	"types": "main.d.ts",
 	"install-peers": "install-peers",
 	"scripts": {
-		"build": "yarn run create-dist && rollup -c",
+		"build": "yarn run create-dist && rollup -c && cp package.json dist/ && cd dist && npm set-script prepare ''",
```
p-forfa, line 10: So the dist/ directory's contents will only exist when you're developing locally now?

p-forfa, line 10: Also, what is the '' at the end of the `npm set-script prepare ''` portion in the build command?

package.json
```diff
@@ -22,8 +22,12 @@
 		"lint:eslint:scripts": "eslint --cache --fix scripts",
 		"lint:eslint:tests": "eslint --cache --fix tests",
 		"lint:eslint": "npm-run-all --continue-on-error --parallel lint:eslint:src",
+		"test": "echo 'No Tests yet'",
 		"prepare": "husky install",
-		"test": "echo 'No Tests yet'"
+		"pub:patch": "npm version patch && rm -rf dist && yarn pub",
```
p-forfa, line 27: Does `patch` automatically bump the "version" in the package.json from say 1.0.0 to 1.0.1, and publish it to artifcatory with that new version?

package.json
JoshuaRogan: Correct

JoshuaRogan: This removes the `husky` command which doesn't work in the dist directory because it's not the root .git directory

package.json
JoshuaRogan: Yup

p-forfa (approved):

PR #89, opened 2022-04-25
DPAD-34 | Add heartbeat tracking

JoshuaRogan:
* Fires events for ads at one second intervals for first ten seconds then one per ten seconds
* Fires event for content at 10 seconds intervals until the first minute than at one minute intervals

### DIFF
--- a/src/utils/videoTracking.ts
+++ b/src/utils/videoTracking.ts
@@ -15,6 +15,8 @@ import {
 	getIsInteractable,
 	getJWAdBlockState,
 	getJWViewState,
+	getIsEmbed,
+	getJWQualityManifest,
 } from 'utils/globalJWInterface';
 import addGlobalProps from 'utils/videoTrackingGlobalProps';
 import getVideoPlayerVersion from 'utils/getVideoPlayerVersion';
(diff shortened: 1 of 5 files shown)


src/utils/videoTracking.ts
```diff
@@ -92,6 +94,8 @@ function addRunTimeParams(trackingParams: TrackData): TrackData {
 	trackDataObject[PROPERTY_NAMES.VIDEO_DURATION] = getDuration();
 	trackDataObject[PROPERTY_NAMES.VIDEO_PLAYER] = getPlayerId();
 	trackDataObject[PROPERTY_NAMES.VIDEO_VIEW_STATE] = getJWViewState();
+	trackDataObject[PROPERTY_NAMES.VIDEO_IS_EMBEDDED] = getIsEmbed();
+	trackDataObject[PROPERTY_NAMES.VIDEO_QUALITY_MANIFEST] = getJWQualityManifest();
```
p-forfa, line 98: So are we storing the available quality levels for the player, whenever it loads?

JoshuaRogan: This is the _possible_ qualities that the video can player. It's available as soon as the player loads. Which all of these events fire after

p-forfa: Just saw it in the doc - https://docs.google.com/spreadsheets/d/1zBUIt8fxmjEItf9HiW9TU1EZnopSLju_TO-BEEM-WuE/edit#gid=1564524057

If I understand it correctly, we'll be sending these the player quality values on each tracking event. The thing I'm uncertain about is whether we'll also send the current quality that the player is watching the video at, not just all the available ones.

p-forfa (approved):

src/utils/videoTracking.ts
JoshuaRogan: Yup, already done via `video_current_quality`

PR #96, opened 2022-04-26
Style fixes article players

gundams-and-dragons:


p-forfa (approved):
If the z-index is over 200, then it should be good.

PR #97, opened 2022-04-27
DPAD-263 :: Temporarily move the setting of JWPlayer version in window object to React component

p-forfa:
The setting of `window.__fandom_jw_player_version` is temporarily being moved to each of the VideoPlayer components, and set through a hook on the initial component render. 

This is not the ideal solution at the moment, since it requires using the hook in each of the VideoPlayer components, but it does solve an issue in UCP. The main.ts script is loaded into the DOM, and causes the player version to be set in the window object, even though the new video player doesn't even load.

Bumped the jwplayer version from "version": "2.0.0-beta-2" to "version": "2.0.0-beta-4", since version 3 was already deployed to artifactory.

### DIFF
--- a/src/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
+++ b/src/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
@@ -10,6 +10,7 @@ import PlayerWrapper from 'players/shared/PlayerWrapper';
 import { Playlist } from 'types';
 import CloseButton from 'players/shared/CloseButton';
 import Attribution from 'players/DesktopArticleVideoPlayer/Attribution';
+import usePlayerVersion from 'hooks/usePlayerVersion';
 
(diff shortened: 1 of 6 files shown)


JoshuaRogan (changes requested):
Why is this code loaded at all when it's not going to be used?

src/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -48,6 +49,7 @@ interface DesktopArticleVideoPlayerProps {
 }
 
 const DesktopArticleVideoPlayer: React.FC<DesktopArticleVideoPlayerProps> = ({ playlist }) => {
+	usePlayerVersion();
```
JoshuaRogan, line 52: We don't need to put this here or the other one. Add it to the SharedJW component. It'll only be once then

PR #99, opened 2022-05-04
DPAD-262 | QA bugs

JoshuaRogan:
* Fix publish date
* Fix mute on volume level
* Append 'jw-' to the video player name
* Add viewability event
* Fix media id for content playing events (required to move the mediaId into the event functions so they trigger at runtime

Remaining Issues: 
* PIP - The events are not properly firing as they should from the JW player API
* video_resume has some additional events that get fired on seek and select next

### DIFF
--- a/src/utils/globalJWInterface.ts
+++ b/src/utils/globalJWInterface.ts
@@ -38,6 +38,10 @@ export const getPlayListId = withTryCatchDefault(() => {
 });
 
 export const getVideoVolume = withTryCatchDefault(() => {
+	if (window.jwplayer().getMute()) {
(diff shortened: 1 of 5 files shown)


src/utils/globalJWInterface.ts
```diff
@@ -38,6 +38,10 @@ export const getPlayListId = withTryCatchDefault(() => {
 });
 
 export const getVideoVolume = withTryCatchDefault(() => {
+	if (window.jwplayer().getMute()) {
+		return 0;
```
p-forfa, line 42: 👍

PR #101, opened 2022-05-18
Update tracking with library updates

JoshuaRogan:
* Use the tracking library to send the events to DW
* Remove some code that is now handled in the library now

### DIFF
--- a/src/utils/globalJWInterface.ts
+++ b/src/utils/globalJWInterface.ts
@@ -56,7 +56,7 @@ export const getCurrentQuality = withTryCatchDefault(() => {
 });
 
 export const getIsCurrentlyViewable = withTryCatchDefault(() => {
-	return !!window.jwplayer().getViewable();
+	return !!window.jwplayer().getViewable() ?? false;
 });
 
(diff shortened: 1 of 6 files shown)


src/utils/globalJWInterface.ts
```diff
@@ -56,7 +56,7 @@ export const getCurrentQuality = withTryCatchDefault(() => {
 });
 
 export const getIsCurrentlyViewable = withTryCatchDefault(() => {
-	return !!window.jwplayer().getViewable();
+	return !!window.jwplayer().getViewable() ?? false;
```
AlburIvan, line 59: Minor nitpick, but wouldn't `??` be redundant here ?

Since the API returns `1` or `0`, `!!` will convert the appropriate value to a boolean and handle `null` and `undefined` accordingly

AlburIvan (approved):
Minor nitpick added, but nonetheless LGTM overall

src/utils/globalJWInterface.ts
JoshuaRogan: Discussed offline and it's some odd issue and this seemed to fix it 🤷

PR #110, opened 2022-07-13
DPAD-786: Twitch: Update UCP UI

gundams-and-dragons:
This PR moves the jwplayer code into a directory, and creates a new directory and player for the twitch player.


src/twitch/players/shared/TwitchPlayerWrapper.tsx
```diff
@@ -0,0 +1,63 @@
...
+`;
+
+const TwitchPlayerTargetWrapper = styled.div`
+	padding-top: 56.25%;
+	position: relative;
+	height: 0;
+`;
+
+interface WindowTwitch extends Window {
+	Twitch?: TwitchApi;
```
p-forfa, line 22: 👍

src/twitch/players/MobileArticleVideoPlayer.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+		if (hasPartnerSlot) {
+			return '75px';
+		}
+
+		return '55px';
+	};
+
+	return (
+		<PlayerWrapper playerName="twitch-mobile-article-video">
```
p-forfa, line 62: Does this use the current player tracking events we have for JW? We might need at least one event to signal that the twitch player was loaded.

src/twitch/players/DesktopArticleVideoPlayer.tsx
```diff
@@ -0,0 +1,92 @@
...
+	top: 0;
+	left: 0;
+	bottom: 0;
+	right: 0;
+	z-index: 2;
+`;
+
+const moveDownAnimation = (right: number, width: number) => keyframes`
+	from {
+		right: ${right}px;
```
AlburIvan, line 24: Can these props be passed down as `undefined` from `placeholderRef.current?.getBoundingClientRect();` ?

p-forfa (approved):
We may want to update the version here. Other than that, looks good.

PR #113, opened 2022-08-09
DPAD-816 :: Add experiment controls

AlburIvan:
## JIRA
[DPAD-816](https://fandom.atlassian.net/browse/DPAD-816)

## Description

Run an A/B test to get a baseline on “what happens to engagement when the video player is not there at all?” 

Target Population
- Desktop & mobile
    - all traffic included, but we should be able to see results per device.

- USA only
    - Since we’re looking at the impact of the video ad experience vs. engagement, a location where we have near-100% sell-through rate is preferred.

- Logged-out traffic only
    - The reason for this is that logged-in folks don’t see wiki-targeted video anyway, so their engagement patterns would skew the test.

- Pageviews containing a video start


src/utils/experiments/checkUserDevice.ts
```diff
@@ -0,0 +1,21 @@
...
+	Unknown = 'unknown',
+}
+
+export default function checkUserDevice(): UserDeviceType {
+	let userDevice = UserDeviceType.Unknown;
+
+	// calculate user device type based on maximum width values per device
+	if (window.matchMedia('(max-width: 767px)').matches) {
+		userDevice = UserDeviceType.Mobile;
+	} else if (window.matchMedia('(min-width: 768px) and (max-width: 991px)').matches) {
```
p-forfa, line 14: Hmm, makes me wonder how this will behave on UCP with the video player. I think at 768px+ its considered desktop. I'm wondering if at these resolutions it WILL load the player.

src/loaders/DesktopArticleVideoLoader.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import JWDesktopArticleVideoPlayer from 'jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer';
+import { DesktopArticleVideoLoaderProps } from 'loaders/types';
+import { jwPlayerExperimentTracker } from 'jwplayer/utils/videoTracking';
+import { checkIfUserInNoVideoExperiment } from 'utils/experiments/experiments';
+
+const DesktopArticleVideoLoader: React.FC<DesktopArticleVideoLoaderProps> = ({ videoDetails }) => {
+	// Disable video load if:
+	if (checkIfUserInNoVideoExperiment()) {
+		jwPlayerExperimentTracker.singleTrack('no-video-experiment');
```
p-forfa, line 10: I'm wondering if we need to differentiate mobile vs desktop in the name here. This might need to be 'desktop-no-video-experiment' and 'mobile-no-video-experiment', unless we're sending extra data in this tracking to differentiate between these.

PR #114, opened 2022-08-12
DPAD-793: Add code splitting to jwplayer-fandom

gundams-and-dragons:


src/loaders/utils/GetVersion.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+declare let window: WindowLibVersion;
+
+export function getVideoPlayerVersion() {
+	return process.env.VIDEO_VERSION;
+}
+
+export function setVersionWindowVar() {
+	window.__fandom_jw_player_version = process.env.VIDEO_VERSION;
```
p-forfa, line 12: Looks like this is only called in useEffects, so it shouldn't be called on the server-side and cause breaks. Just to be future proof, I would add a window undefined check. But for now, I think this should be ok.

p-forfa (approved):
LGTM. Added one comment about server-side checks on window variable access.

PR #115, opened 2022-08-23
DPAD-1030 :: Added class names to the mobile player elements to fix ToC overlap issue

p-forfa:
* Added class names to the mobile player elements to fix ToC overlap issue
* The code in UCP responsible for shifting the ToC based on certain class names is found in the following two places:
* https://github.com/Wikia/unified-platform/blob/bb5a7385fd77c1da778a45e24480848bc35f56f0/skins/FandomMobile/scripts/TableOfContents.ts#L35
* https://github.com/Wikia/unified-platform/blob/bb5a7385fd77c1da778a45e24480848bc35f56f0/skins/FandomMobile/scripts/TableOfContents.ts#L64
* There is still a small issue with the ToC element overlapping the right side of the bottom title bar, but this PR addresses the main issue of the ToC overlapping the close button
* Added classnames on certain video player elements, so that regression tests can target them easily

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@fandom/jwplayer-fandom",
-	"version": "2.0.0-beta-18",
+	"version": "2.0.0-beta-21",
 	"description": "JWPlayer for Fandom",
 	"main": "bundle.umd.js",
 	"module": "bundle.esm.js",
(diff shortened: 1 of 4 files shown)


JoshuaRogan (approved):
Lgtm but I prefer data attributes over class names for this because it makes it clear it shouldn't be touched

PR #116, opened 2022-09-12
DPAD-1068 :: Move Close button down on Mobile Ad

p-forfa:
* The close button needs to be moved down, below the Learn More overlay, whenever an ad plays
* Added in a react hook to check whether the player is in an ad or not
* The close button is moved down by a static amount of 40px. Grabbing the Learn More button element is slightly difficult, given that the element exists inside of an iFrame that's sourced from a different domain

### DIFF
--- a/src/utils/useAdBreak.ts
+++ b/src/utils/useAdBreak.ts
@@ -1,0 +1,38 @@
+import { useState, useEffect, useContext } from 'react';
+import { PlayerContext } from 'jwplayer/players/shared/PlayerContext';
+import JWEvents from 'jwplayer/players/shared/JWEvents';
+
+export default function useAdBreak(): boolean {
+	const [adBreak, setAdBreak] = useState(false);
+	const { player } = useContext(PlayerContext);
+
+	useEffect(() => {
+		player?.on(JWEvents.AD_BREAK_START, () => {
+			console.log('AD_BREAK_START');
+			setAdBreak(true);
+		});
+		player?.on(JWEvents.AD_STARTED, () => {
+			console.log('AD_STARTED');
+			setAdBreak(true);
+		});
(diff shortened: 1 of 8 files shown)


src/utils/useAdBreak.ts
```diff
@@ -0,0 +1,38 @@
...
+import { PlayerContext } from 'jwplayer/players/shared/PlayerContext';
+import JWEvents from 'jwplayer/players/shared/JWEvents';
+
+export default function useAdBreak(): boolean {
+	const [adBreak, setAdBreak] = useState(false);
+	const { player } = useContext(PlayerContext);
+
+	useEffect(() => {
+		player?.on(JWEvents.AD_BREAK_START, () => {
+			console.log('AD_BREAK_START');
```
p-forfa, line 11: This ad_break_start event does not seem to fire whenever the ad plays

src/utils/useAdBreak.ts
```diff
@@ -0,0 +1,38 @@
...
+			console.log('AD_LOADED');
+			setAdBreak(true);
+		});
+		player?.on(JWEvents.AD_PLAY, () => {
+			console.log('AD_PLAY');
+			setAdBreak(true);
+		});
+
+		player?.on(JWEvents.AD_BREAK_END, () => {
+			console.log('AD_BREAK_END');
```
p-forfa, line 28: Same thing with this event.

src/utils/useAdBreak.ts
gundams-and-dragons: We may not be triggering this with the way we're loading ads.

p-forfa: Got it. Maybe that'll get fixed if we utilize the built in JW API ad methods sometime in the future.

AlburIvan (approved):
lgtm

PR #117, opened 2022-09-14
Pause player fix

gundams-and-dragons:


src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -73,6 +72,12 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({ config, playerUrl, on
 				}
 			});
 
+			playerInstance.on(JWEvents.PLAYLIST_ITEM, (event: OnPlaylistItemEventData) => {
+				if (event.index === 2) {
+					playerInstance.pause();
```
JoshuaRogan, line 77: This makes sense in theory, I'm just slightly worried running the 'pause' command can have some issues with tracking or may introduce something unexpected. I say we ship and monitor that after we launch though.

PR #118, opened 2022-09-15
Make sure all videos after the third video (including the third one),…

p-forfa:
… always pause before the video starts (don't autoplay) | Increase package version for release


AlburIvan (approved):
lgtm

PR #120, opened 2022-09-16
DPAD-1085 :: Fix canonical player not reloading properly

AlburIvan:
### Jira

[DPAD-1085 :: Video player doesn't refresh after clicking on a different video](https://fandom.atlassian.net/browse/DPAD-1085)

### Description

When I click on a video from “Videos you might like”, the player still plays the same video. New video appears only after I refresh the page.

Fixed this behavior caused by the lazy loader ignoring props change on SPAs

- [x] QA Approved

### Test links

[FAT Deployment](https://build-dpad-1085.apps.fandom-dev.us/video/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13620 @@
+{
+	"name": "@fandom/jwplayer-fandom",
+	"version": "2.0.0-beta-25",
+	"lockfileVersion": 2,
+	"requires": true,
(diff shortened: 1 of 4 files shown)


p-forfa (approved):
LGTM. Tested on https://build-dpad-1085.apps.fandom-dev.us/video as well.

PR #122, opened 2022-09-16
Put in a fix to not render the Desktop and Mobile UCP players when th…

p-forfa:
…e window.canPlayVideo() function returns false; this includes guards in case the functions is undefined, or even if the window variable is not defined for whatever reason, including SSR

### DIFF
--- a/src/loaders/DesktopArticleVideoLoader.tsx
+++ b/src/loaders/DesktopArticleVideoLoader.tsx
@@ -6,14 +6,10 @@ import { shouldLoadUcpPlayer } from 'loaders/utils/shouldLoadPlayer';
 export { getVideoPlayerVersion } from 'loaders/utils/GetVersion';
 
 export const DesktopArticleVideoLoader: React.FC<DesktopArticleVideoLoaderProps> = ({ videoDetails }) => {
-	if (!shouldLoadUcpPlayer()) {
-		return;
-	}
-
 	const [player, setPlayer] = useState(undefined);
 
 	useEffect(() => {
-		if (!player) {
+		if (!player && shouldLoadUcpPlayer()) {
 			getPlayer();
 		}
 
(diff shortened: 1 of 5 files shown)


src/loaders/DesktopArticleVideoLoader.tsx
```diff
@@ -1,10 +1,15 @@
...
 import { setVersionWindowVar } from 'loaders/utils/GetVersion';
+import { shouldLoadUcpPlayer } from 'loaders/utils/shouldLoadPlayer';
 
 export { getVideoPlayerVersion } from 'loaders/utils/GetVersion';
 
 export const DesktopArticleVideoLoader: React.FC<DesktopArticleVideoLoaderProps> = ({ videoDetails }) => {
+	if (!shouldLoadUcpPlayer()) {
+		return;
+	}
+
```
AlburIvan, line 12: Quick question, can one do that? 

Won't it have an issue with `hooks` cannot be called conditionally ?

gundams-and-dragons: I think the hooks will get called either way, but I'd move this below those, just above the other returns

gundams-and-dragons: Actually inside the effect hook would probably be best, prevent it from getting run altogether

gundams-and-dragons: Maybe something like

if (!player && shouldLoadUcpPlayer()) {
	getPlayer();
}

p-forfa: I'll make the change. There's a deploy running on sandbox-s8 right now. I'll verify it works then, then redeploy with the above change you both discussed.

p-forfa: @gundams-and-dragons @AlburIvan Fixed in 4398dd619dc73b31e4c4b49a8e1ef58f2f0e5a85

gundams-and-dragons (approved):

AlburIvan (approved):

src/loaders/DesktopArticleVideoLoader.tsx
JoshuaRogan: @gundams-and-dragons is `canPlayVideo` created from the php side too? If not the hook could be an issue if our code somehow runs first.

JoshuaRogan: Just checked warrens slack. LGTM

PR #130, opened 2022-11-08
DPAD-1213 :: Stop auto-advancement if the player is not viewable

AlburIvan:
### Issue

[DPAD-1213](https://fandom.atlassian.net/browse/DPAD-1213)

### Description

JW has a built-in option to prevent auto-starting until the player is in “view” (“viewable” from an advertiser perspective, >50% of pixels). However, this setting only affects the initial play. If “related” plays continue afterward, they will start regardless of the player’s visibility.

This PR addresses this issue by counting the number of videos that has been watched and then stopping the player if its not on screen (at least 50% on the screen according to JW).

#### Why an external index instead of JW's `index` on `playlistItem` callback ?

Simply because that index changes when another playlist is selected (be it manually or automatically), so its not reliable enough to be used to count watched videos.

### Links

[sandbox-s7](https://paweltest2.sandbox-s7.fandom.com/wiki/Test-video)

### DIFF
--- a/src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
+++ b/src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
@@ -98,7 +98,7 @@ const DesktopArticleVideoPlayer: React.FC<DesktopArticleVideoPlayerProps> = ({ v
 						<JwPlayerWrapper
 							config={getArticleVideoConfig(videoDetails)}
 							onReady={(playerInstance) => articlePlayerOnReady(videoDetails, playerInstance)}
-							pauseOnExitViewport={true}
+							stopAutoAdvanceOnExitViewport={true}
 						/>
 						{isScrollPlayer && <VideoDetails />}
 					</DesktopArticleVideoWrapper>
--- a/src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
+++ b/src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
@@ -98,6 +98,7 @@ const DesktopArticleVideoPlayer: React.FC<DesktopArticleVideoPlayerProps> = ({ v
 						<JwPlayerWrapper
 							config={getArticleVideoConfig(videoDetails)}
 							onReady={(playerInstance) => articlePlayerOnReady(videoDetails, playerInstance)}
+							pauseOnExitViewport={true}
 						/>
 						{isScrollPlayer && <VideoDetails />}
 					</DesktopArticleVideoWrapper>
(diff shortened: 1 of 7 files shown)


src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -98,6 +98,7 @@ const DesktopArticleVideoPlayer: React.FC<DesktopArticleVideoPlayerProps> = ({ v
 						<JwPlayerWrapper
 							config={getArticleVideoConfig(videoDetails)}
 							onReady={(playerInstance) => articlePlayerOnReady(videoDetails, playerInstance)}
+							pauseOnExitViewport={true}
```
JoshuaRogan, line 101: This variable makes it sound like it will always pause when it exits the viewport. Which is also a reasonable feature we might do in the future (or at least a/b test). Can we rename it to something more precise?

JoshuaRogan (approved):
Minor comment but code LGTM

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
p-forfa: stopAutoAdvanceOnExitViewport might work here.

p-forfa: Actually, looking at the behavior the puaseOnExitViewport makes sense. It looks like the player advances to the next video, and then stops/pauses before the video plays. If the user is on another tab while the next video loads, it just pauses.

I think @AlburIvan mentioned that the JW Player behavior for stopping auto advancing only worked on the initial video. This is probably the best compromise we have at the moment to prevent the video from actually playing.

I am curious how this works with ads though. I haven't gotten one on sandbox. My assumption is that the PLAYLIST_ITEM callback gets called when the actual video is about to play, not the ad.

p-forfa (approved):
The changes LGTM. We'll need to verify the behavior with Jase Sanders. Given the 'auto advance off' limitations (only stops auto advance after the first video play) that were found with the JW Player, this seems like a good start/compromise to at least get something out.

src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -87,6 +92,26 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({ config, playerUrl, on
...
+				}
+
+				// if the video is on its 2nd+ play, pause the video if its not on the viewport
+				if (videoIndexRef.current >= 1 && (playerInstance.getViewable() === 0 || document.hasFocus() === false)) {
+					// send tracking event
+					jwPlayerPlaybackTracker({ event_name: 'video_player_pause_not_viewable' });
+
+					// close the related UI to stop auto advancement, and then re-open it for users to click on
+					setTimeout(() => {
+						window.jwplayer(elementId).getPlugin('related').close();
```
p-forfa, line 111: Nice workaround. Good to know about getPlugin.

p-forfa (approved):
LGTM after adding in the getPlugin workaround to turn off/on the UI. Tested in on the linked sandbox-s7 link and it seems to stay there after I go to another tab.

p-forfa (approved):

PR #131, opened 2022-11-10
Dpad-995 :: experiment logic test for Desktop Article Video Loader

p-forfa:
* Added in the pathfinder-lite library to take advantage of the bucketing logic
* Added in buckets p and q as test buckets (subject to change)
* Added in pathfinder-lite bucketing logic to ensure that the reskinned player will load

Notes:
The query string `pf_pathfinder_force_bucket=p` or `pf_pathfinder_force_bucket=q` can be used to force the experiment, by providing this query param on the browser.  P and Q buckets can be replaced with any other buckets, as long as they're reflected in the defineExperiment function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
 		"@babel/preset-env": "^7.16.5",
 		"@babel/preset-react": "^7.16.5",
 		"@babel/preset-typescript": "^7.16.5",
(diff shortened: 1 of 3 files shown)


AlburIvan (approved):
LGTM

PR #139, opened 2022-12-19
limiting resolution of related videos

gundams-and-dragons:


src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -60,6 +61,9 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 		jwPlayerPlaybackTracker({ event_name: 'video_player_start_load' });
 
 		const onload = () => {
+			// Set the max_resolution param for related videos
+			window.jwplayer.defaults.related.file = window.jwplayer.defaults.related.file + '&max_resolution=1280';
```
JoshuaRogan, line 65: Can we ensure this variable is there?

gundams-and-dragons: I believe so, but we can put a check in

p-forfa (approved):

PR #149, opened 2023-02-07
DPAD-1485 :: Restructure the RedVenture stand-alone player bundling

p-forfa:
-Restructured the compilation process of the stand-alone video player
-The stand-alone dist now has its own package.json file, and functions as a 'stand-alone app'. This approach helps with changing the peerDependencies in the main package.json, where previously the react, react-dom and other libraries had to be added in as devDependencies in order to have the stand-alone player import the necessary libraries to run by itself.
-Added in the RedVentureVideoPlayer as a separate export in the main package.json and in the main rollup.config.js file, so that the stand-alone app can import it. Using the RedVenture player export provided by the main.js file in the dist directory caused all the other players (Desktop, Mobile and Canonical players) to be bundled into the standAlone_RV_VideoPlayer.js package and increasing the bundle size by over 100kb. A direct file reference prevent the bundle size from increasing.


stand-alone/package.json
```diff
@@ -0,0 +1,65 @@
+{
+	"name": "@fandom/rv-jwplayer",
+	"version": "2.1.5",
```
gundams-and-dragons, line 3: The version and description should probably be changes to reflect this app instead of the library.

gundams-and-dragons (changes requested):

stand-alone/package.json
p-forfa: Fixed in 11b086d1be3a655e8ad24d9eceaff73ae0f51813

gundams-and-dragons (approved):

PR #155, opened 2023-02-13
DPAD-1436: Add a outline or border to the close button so that it's visible on white backgrounds

gundams-and-dragons:


p-forfa (approved):
LGTM

PR #159, opened 2023-02-27
VIDEO-194 Default player for 5 N&R sites

nandy-andy:


src/youtube/players/shared/youtubeTrackingEvents.ts
```diff
@@ -1,5 +1,5 @@
 import { track } from '@fandom/tracking-metrics/tracking/dataLayer';
-import { YoutubeTakeOverDetails } from 'loaders/utils/GetYoutubeTakeoverDetails';
+import { YoutubeTakeOverDetails } from 'youtube/types';
```
nandy-andy, line 2: @p-forfa It removes problems while building in the console. Am I right thinking it shouldn't break anything? How can I double-check it? ;-)

p-forfa: Ahh, good catch. I'm not sure why that wasn't flagged by my IDE. By running 'yarn build', it should be enough since this is a typescript definition.

The other good way to actually check on the player itself, is to import the the 'DesktopArticleVideoLoader' component inside /test-jw/src/App.js, and run 'yarn app' from the source directory. 

It should open up a localhost page. Add this query param `?youtube_embed_video_id=F0Ga_nPZuiI` where `F0Ga_nPZuiI` is some youtube video id. This will have the Youtube Video Embed takeover. 

I'll document the local Fandom JWPlayer process in the README so its clearer.

stand-alone/loaderHelper.ts
```diff
@@ -184,3 +184,14 @@ export const canPlayerRender = (context: RedVenturePlayerContextProps): boolean
 
 	return true;
 };
+
+const newsAndRatingsNamesWithDefaultPlayer = [ 'gamespot', 'gamefaqs', 'comicvine', 'tvguide', 'metacritic' ];
+export const assurePlayerUrl = (context: RedVenturePlayerContextProps): RedVenturePlayerContextProps => {
+	const newContext = { ...context };
```
p-forfa, line 190: Makes sense to me. We could do the copy in the if statement, and return the new object with the newly set playerUrl only if we get to the `if` statement. Otherwise, we can just return the old untouched context object, in cases of 'giantbomb' for example.

I'm good with the approach you have as well 👍

nandy-andy: He he, that was my original code and I decided to change it since IMHO it was less readable 😊

src/youtube/players/shared/youtubeTrackingEvents.ts
nandy-andy: Seems working 👍

p-forfa (approved):
LGTM. I'd do a deployment to the test env with `yarn deploy-stand-alone --env test` and check it out through https://static.wikia.nocookie.net/silversurfer/video/test/standalone-dist/standAlone_RV_VideoPlayer.js

Tomasz-at-Fandom (approved):

PR #163, opened 2023-03-22
VIDEO-454 Remove unnecessary console.logs

nandy-andy:


rafalgiemza (commented):
I agree in general to cut those lines. 
Just wonder what if - instead of removing - hide them until someone enables a flag in the URL like `?video_debug=true`.

nandy-andy:
That's a good idea, probably useful for debugging. However, I wouldn't block this changes as now without someone willing the console is flooded once an ad is served in JWPlayer.

Let's discuss it more on a refinement because there is some logging logic in the video player and also we could ad logging for video ads in the AdEngine. So, it's a matter of deciding and unifying the logging ;) Feel free to fill a ticket.

PR #164, opened 2023-04-19
Ad skip fix for gamespot on mediaId HmjhpQRV

p-forfa:
For reference - not an actual fix. This has been deployed to Prod on April 19th, 2023 to fix an issue where ads should not play on a page where HmjhpQRV mediaId is playing on Gamespot


nandy-andy:
Thanks! We'll address it in more general fix: https://github.com/Wikia/jwplayer-fandom/pull/168

PR #168, opened 2023-05-10
VIDEO-527 Request article-video service for sponsored videos list

nandy-andy:


kwfandom (approved):

src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -46,9 +46,15 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 	useEffect(() => {
 		if (shouldLoadSponsoredContentList) {
 			const retrieveSponsoredVideo = async () => {
```
pbartkowiak-fandom, line 48: Why defining a function inside the `useEffect` hook? The hook could be leaner without it.

pbartkowiak-fandom (approved):

src/jwplayer/players/shared/JwPlayerWrapper.tsx
nandy-andy: I agree. We could extract it and even unit-test it. I'll fill a follow-up ticket, so we don't slow down the process as it's needed to be live on Monday 🤷‍♂️

rafalgiemza (approved):

src/jwplayer/players/shared/JwPlayerWrapper.tsx
nandy-andy: https://fandom.atlassian.net/browse/VIDEO-580

PR #176, opened 2023-08-01
Prevent view from scrolling to the top between videos

pbartkowiak-fandom:
https://fandom.atlassian.net/browse/COTECH-670


src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
```diff
@@ -55,6 +55,7 @@ export const MobileArticleVideoPlayerContent: React.FC<MobileArticleVideoPlayerP
 	const [dismissed, setDismissed] = useState(false);
 	const isScrollPlayer = !(dismissed || onScreen);
 	const inputName = 'isDismissed';
+	const relatedContainer: HTMLDivElement | null = document.querySelector('#featured-video__player_related');
```
nandy-andy, line 58: Since we're selecting an element with unique id what do you think of just using `document.getElementById()`? 🤔

test-jw/src/app.css
```diff
@@ -7,4 +7,8 @@
 
 body {
     height: 200vh;
-}
\ No newline at end of file
+}
+
+.hidden {
+	display: none !important;
```
nandy-andy, line 13: Does it work without `!important` or gets overwritten? 😊

test-jw/src/app.css
```diff
@@ -7,4 +7,8 @@
 
 body {
     height: 200vh;
-}
\ No newline at end of file
+}
+
+.hidden {
```
nandy-andy, line 12: Consider changing the class name to `forcefullyHidden` or `fandomHidden`, so it's less unlikely it'll get overwritten by our users/wikis custom CSS. Recently, there was an issue in Ad Engineering that the custom CSS was overwritting AdEngin's `.hidden` class - let's learn from it and as long as we allow custom CSS let's avoid common CSS class names.

src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
```diff
@@ -76,9 +77,12 @@ export const MobileArticleVideoPlayerContent: React.FC<MobileArticleVideoPlayerP
 
 	useEffect(() => {
 		if (!onScreen) {
+			relatedContainer?.classList.add('hidden');
```
nandy-andy, line 80: Maybe another way to achieve our goal is to close the related plugin when it's not in the viewport 🤔 https://docs.jwplayer.com/players/reference/related

test-jw/src/app.css
pbartkowiak-fandom: The `!important` is crucial for this to work!

src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
pbartkowiak-fandom: 🤷 There's hardly a difference. I find it more consistent to use the `querySelector` most of  the time instead of switching between different selectors.

test-jw/src/app.css
pbartkowiak-fandom: Agreed 👍

src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
pbartkowiak-fandom: Yes, but I guess we want to maintain the _Related Videos_ feature.

src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
nandy-andy: But there is a difference and it's connected to speed, so maybe it's worth ;-) 

Actually, if there are more than one JWP on page there would be more `.jw-plugin-related` elements, so using `querySelector()` seems like a better idea. But then using the ID in the selector doesn't make sense and we'd need a selector with the player ID and the plugin class, for example: `#featured-video__player-container .jw-plugin-related`.

test-jw/src/app.css
nandy-andy: So, it gets overwritten. But it's a bad practice  😿 Thus why I started looking for other solutions 🤔

pbartkowiak-fandom: Hmm, we can avoid setting the _!important_ class by setting the inline style in .jsx component instead. I can give it a go 👍

nandy-andy (approved):

PR #180, opened 2023-08-10
FCPPT-25 performance improvements

David-Iacovone:
Tickets: 

- https://fandom.atlassian.net/browse/FCPPT-89
- https://fandom.atlassian.net/browse/FCPPT-41
- https://fandom.atlassian.net/browse/FCPPT-25

This PR includes the following changes:

- Communication service update
- Legacy test code removal
- Updating old dependencies
- Github workflows update
- Added a loading screen while video is loading
- Prevented JW Player from downloading poster image for videos


houssem-yahiaoui:
@David-Iacovone can we remove RxJs package since we're not using it in the communicationservice ?

sSwiergosz (approved):

src/jwplayer/utils/useAdComplete.ts
```diff
@@ -1,16 +1,16 @@
...
 
 export default function useAdComplete(): boolean {
 	const [adComplete, setAdComplete] = useState(false);
+	const communicationService = getCommunicationService();
 
 	useEffect(() => {
 		recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_AD_ENG_OPT_IN_LISTEN_START);
-		communicationService.action$.pipe(ofType('[AdEngine OptIn] set opt in'), first()).subscribe(() => {
+		communicationService.on('[AdEngine OptIn] set opt in', () => {
+			console.log(' ---- [AdEngine OptIn] set opt in ---- ');
```
nandy-andy, line 13: Let's remove `console.log()`, so we don't add noise in the dev console. Or replace it with `console.debug()` so it doesn't show up in a regular dev tools console (only when someone filters for all the messages). Ultimately, I'd like to add there one more layer and push debug info only when we have a cookie or query-string 🤔 But `console.log()` is a no-no from me ;-)

src/jwplayer/utils/useAdComplete.ts
```diff
@@ -25,17 +25,21 @@ export default function useAdComplete(): boolean {
...
-		return race(adEngineConfigured$, adEngineTimeout$).toPromise();
+		return Promise.race([adEngineConfigured$, adEngineTimeout$]);
 	};
 
 	const listenSetupJWPlayer = (callback) => {
-		communicationService.action$.pipe(ofType('[Ad Engine] Setup JWPlayer'), first()).subscribe(callback);
+		communicationService.on('[Ad Engine] Setup JWPlayer', () => {
+			callback();
+		});
+		console.log(' ---- [Ad Engine] Setup JWPlayer ---- ');
```
nandy-andy, line 42: Similarly to the comment above: let's replace it for now with `console.debug()`. Also, I wonder if it shouldn't be in the callback in `communicationService.on()` above 🤔 It depends what you wanted to signal with the log -- the fact we hooked to the event or the fact the event happened and we reacted on it calling the callback?

src/jwplayer/utils/useAdComplete.ts
houssem-yahiaoui: A Debug query string is a good idea

agatakozinska (approved):

nandy-andy (approved):

agatakozinska:
@David-Iacovone could you take care of the checks so we could test it?

latata (commented):
Can we remove `experimental` directory?

package.json
```diff
@@ -132,13 +133,12 @@
 		"react-i18next": "^11.7.1"
 	},
 	"dependencies": {
-		"@fandom-frontend/design-system": "15.18.1",
-		"@fandom-frontend/react-common": "13.23.1",
+		"@fandom-frontend/design-system": "15.13.1",
+		"@fandom-frontend/react-common": "13.19.0",
```
nandy-andy, line 137: this is interesting because it looks like a downgrade 🤔

David-Iacovone: Well, I've been updating from version 9 to 15 for example. A new version must have been added during my PTO

David-Iacovone: It's fixed

PR #182, opened 2023-08-16
COTECH-683 (fix): handle missing duration property

sSwiergosz:
[Ticket](https://fandom.atlassian.net/browse/COTECH-683)

I noticed that for some videos (for example first video on the Yoda page) instead of the video duration there is `NaN:NaN` displayed. Currently, there is an if statement that checks if property `duration` is less or equal to zero but it doesn't take into account situations when duration is missing (it's the case for Yoda's video).

Also, it's better to move this statement as a first instruction in the function - we don't want to compute anything if the duration is not a number.

cc @Wikia/site-experience @Wikia/cotech

### DIFF
--- a/src/jwplayer/utils/formatTime.ts
+++ b/src/jwplayer/utils/formatTime.ts
@@ -1,15 +1,19 @@
+function isNumber(value: unknown): boolean {
+	return typeof value === 'number' && !isNaN(value);
+}
+
 export function formatTime(duration: number): string {
+	// If duration is below 0 seconds or is not a number it means corrupted data, we don't want to display it
+	// Also return early for 0 seconds
+	if (duration <= 0 || !isNumber(duration)) {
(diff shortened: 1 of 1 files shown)


dzawada (approved):

sSwiergosz:
> It makes sense to check it earlier and exit the function sooner if there is no duration.
> 
> However, these change are only hiding the real bug. Because AFAIK the bug appears only on Fandom Mobile when scrolling and affects all the videos not only Yoda one (which clearly has its duration in the JWP dashboard: 1:38).
> 
> I'd merge the PR because of moving the check earlier but I'll look for the root cause within the COTECH ticket we have in our backlog ;-) Thanks!

@nandy-andy That's true, it only hides the real problem but since it's a utility function that might be used in many places it's good to have that basic type guard :)

PR #186, opened 2023-09-04
FCPPT-103: Fix jwplayer placeholder issues

agatakozinska:
https://fandom.atlassian.net/browse/FCPPT-103

Deployed: https://starwars.sandbox-xw1.fandom.com/wiki/Yoda

Description (extended from ticket description):
- The timing for showing placeholder was changes, we look now for readiness of player not `adComplete` particularly because this variable name is misleading - it triggers right after receiving jwplayer setup but jwplayer is still not rendered so placeholder should be still there in this case.
- Video details were shown before even loading video player which is why it also should wait for video player.
- Shadow was moved to the video details to not blend with fandom page.

<details>
<summary>After (sandbox on the left) vs. before (production on the right)</summary>

https://github.com/Wikia/jwplayer-fandom/assets/25872895/f733b496-9b71-4cd8-906d-14a37e8abf6b

https://github.com/Wikia/jwplayer-fandom/assets/25872895/3bbf48ff-e6fd-4e69-982e-64b1b053cda7
</details>

### DIFF
--- a/src/jwplayer/players/MobileArticleVideoPlayer/mobileArticleVideoPlayer.module.scss
+++ b/src/jwplayer/players/MobileArticleVideoPlayer/mobileArticleVideoPlayer.module.scss
@@ -1,9 +1,10 @@
 .mobileArticleVideoTopPlaceholder {
+	--z1: token(z1);
 	width: 100%;
 	height: 56.25vw;
 	position: relative;
-	// must be bigger than infobox button
-	z-index: 101;
+	// z-index must be bigger than infobox button
+	z-index: calc(var(--z1 + 1);
 }
 
 .mobileArticleVideoWrapperIsScrollPlayer {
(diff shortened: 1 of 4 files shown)


src/jwplayer/players/MobileArticleVideoPlayer/mobileArticleVideoPlayer.module.scss
```diff
@@ -2,12 +2,12 @@
 	width: 100%;
 	height: 56.25vw;
 	position: relative;
-	z-index: 100;
+	// must be bigger than infobox button
+	z-index: 101;
```
nandy-andy, line 6: Let's use a variable if possible instead of a hardcoded value with a comment 🤔

nandy-andy (approved):

houssem-yahiaoui (approved):

sSwiergosz (approved):

PR #187, opened 2023-09-05
COTECH-702 Fix CanonicalVideoPlayer

nandy-andy:


andrzejkupczyk (approved):

src/jwplayer/players/CanonicalVideoPlayer/CanonicalVideoPlayer.tsx
```diff
@@ -6,22 +6,8 @@ import useAdComplete from 'jwplayer/utils/useAdComplete';
 import { communicationService } from 'jwplayer/utils/communication';
 import { isLocalDevelopment, isOnBrowser } from 'jwplayer/utils/envs';
 
-import clsx from 'clsx';
-
 import styles from './canonicalVideoPlayer.module.scss';
```
andrzejkupczyk, line 9: Is this import still required?

andrzejkupczyk: Ok, now I see. It is still being used.

src/jwplayer/utils/useAdComplete.ts
```diff
@@ -4,10 +4,16 @@ import { communicationService, ofType } from 'jwplayer/utils/communication';
 import { race, timer } from 'rxjs';
 import { first } from 'rxjs/operators';
 import { recordVideoEvent, VIDEO_RECORD_EVENTS } from 'jwplayer/utils/videoTimingEvents';
+import getValueFromQuery from 'utils/getValuefromQuery';
 
 export default function useAdComplete(): boolean {
 	const [adComplete, setAdComplete] = useState(false);
 
+	const noAds = getValueFromQuery('noads');
+	if (noAds) {
```
latata, line 13: It's not recommended to have such if statements with return before hooks (`useEffect` will run conditionally). In this case, it probably won't break anything as `noads` will alwasy be there through the whole react app lifecycle but I think we should still try to avoid such things in the code

nandy-andy: Since it's not really working locally, I'm fine with removing it 😊 Thanks 🙇

PR #189, opened 2023-09-06
COTECH-683 | Make playlistItem undefined by default, fix typescript issues

latata:
https://fandom.atlassian.net/browse/COTECH-683

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,5 +140,8 @@
 		"clsx": "1.2.1",
 		"js-cookie": "3.0.1",
 		"rxjs": "7.5.4"
+	},
+	"resolutions": {
+		"**/@types/react": "17.0.39"
 	}
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -140,5 +140,8 @@
 		"clsx": "1.2.1",
 		"js-cookie": "3.0.1",
 		"rxjs": "7.5.4"
+	},
+	"resolutions": {
+		"**/@types/react": "17.0.39"
```
latata, line 145: I noticed that when running `tsc --noEmit` there were a lot of TS errors reported. I was caused by multiple instances of `@types/react` in node_modules. This line fixes the issue.

src/jwplayer/utils/usePlaylistItem.ts
```diff
@@ -2,9 +2,9 @@ import { useContext, useState, useEffect } from 'react';
 import { PlaylistItem, PlaylistItemPlayerEventData } from 'jwplayer/types';
 import { PlayerContext } from 'jwplayer/players/shared/PlayerContext';
 
-export default function usePlaylistItem(): PlaylistItem {
+export default function usePlaylistItem() {
 	const { player } = useContext(PlayerContext);
-	const [playlistItem, setPlaylistItem] = useState<PlaylistItem>({ title: undefined, duration: undefined });
+	const [playlistItem, setPlaylistItem] = useState<PlaylistItem>();
```
latata, line 7: Make `playlistItem` undefined by default. VideoDetails will not render NaN:NaN then.

PR #193, opened 2023-09-08
COTECH-711 | Typescript check

latata:
https://fandom.atlassian.net/browse/COTECH-711

### DIFF
--- a/JenkinsFile
+++ b/JenkinsFile
@@ -59,7 +59,7 @@ pipeline {
           steps {
             sh("yarn tsc --noEmit")
             script {
-              FAILED_STAGE = 'lint'
+              FAILED_STAGE = 'typescript'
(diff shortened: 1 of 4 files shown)


JenkinsFile
```diff
@@ -55,6 +55,14 @@ pipeline {
           }
         }
 
+        stage('typescript') {
+          steps {
+            sh("yarn tsc --noEmit")
+            script {
+              FAILED_STAGE = 'lint'
```
andrzejkupczyk, line 62: Should we add this to the existing lint stage, or update the FAILED_STAGE value to the correct one?

latata: good catch I will change it to `typescript`

nandy-andy (approved):

PR #194, opened 2023-09-11
COTECH-698 | artifactory cleanup script

latata:


scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,69 @@
...
+			},
+		);
+	});
+}
+
+function extractVersionFromPackageName(packageName) {
+	return packageName.match(/^\/jwplayer-fandom-(\d+\.\d+\.\d+).*\.tgz/)?.[1];
+}
+
+function isTestVersion(packageName) {
```
latata, line 33: test version = every version that is not valid semver (`X.Y.Z`)

scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,69 @@
...
+const packageJson = require('../package.json');
+
+const BASE_ARTIFATORY_URL = 'https://artifactory.wikia-inc.com/artifactory';
+const BASE_PACKAGE_PATH = '/npm-local/@fandom/jwplayer-fandom/-/@fandom';
+const BASE_READ_PATH = `/api/storage${BASE_PACKAGE_PATH}`;
+const BASE_DELETE_PATH = BASE_PACKAGE_PATH;
+
+const currentMajorVersion = semver.major(packageJson.version);
+const previousMinorVersion = semver.minor(packageJson.version) - 1;
+const versionThreshold = `${currentMajorVersion}.${previousMinorVersion}.0`;
```
latata, line 12: let's remove all test versions lower than current version - 1 minor version:
e.g. if the current version in `2.6.4` we will remove all test packages that are marked as `2.4.X` and lower (we'll keep `2.6.X` and `2.5.X`).

scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,69 @@
...
+async function deleteTestPackagesWithVersionLowerThan(version) {
+	const packages = await fetchTestPackages();
+	const packagesToDelete = packages.filter((packageName) => {
+		const packageVersion = extractVersionFromPackageName(packageName);
+		return !packageVersion || semver.lt(packageVersion, version);
+	});
+
+	for (const packageName of packagesToDelete) {
+		console.log(`Deleting ${packageName}...`);
+		// await deletePackage(packageName);
```
latata, line 65: I will uncomment it ;)

scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,69 @@
...
+const BASE_PACKAGE_PATH = '/npm-local/@fandom/jwplayer-fandom/-/@fandom';
+const BASE_READ_PATH = `/api/storage${BASE_PACKAGE_PATH}`;
+const BASE_DELETE_PATH = BASE_PACKAGE_PATH;
+
+const currentMajorVersion = semver.major(packageJson.version);
+const previousMinorVersion = semver.minor(packageJson.version) - 1;
+const versionThreshold = `${currentMajorVersion}.${previousMinorVersion}.0`;
+
+async function getArtifactoryTokenFromYarnConfig() {
+	return new Promise((resolve, reject) => {
```
andrzejkupczyk, line 15: When you use async in front of a function, it automatically wraps the returned value in a Promise. So if you explicitly return a new Promise inside an async function, you'd be wrapping a Promise inside another Promise, which is usually unnecessary and can complicate your code.

scripts/artifactory-cleanup.js
andrzejkupczyk: This comment could be a part of a code documentation 😉 It might be a good idea to copy this detail into a jsdoc.

scripts/artifactory-cleanup.js
andrzejkupczyk: Should we deprecate versions before removing them? npm has a command for this (`npm deprecate <package-name>@<version> "<message>"`), not sure about Yarn.

latata: I think it's not needed as we are going to remove them sooner rather than later. No one will notice that these packages are deprecated as no one is using them anymore.

andrzejkupczyk: Okay, I wasn't sure about their current usage. According to @nandy-andy, we don't have a full list of dependents at the moment.

David-Iacovone: That's true. However, it's very likely that the only consumers here are UCP and NAR

scripts/artifactory-cleanup.js
David-Iacovone, line 65:
```suggestion
		await deletePackage(packageName);
```

David-Iacovone (changes requested):
I would suggest to update general README about this new feature. Other than that it looks really good.

David-Iacovone (approved):

scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * USAGE:
+ *
+ * node scripts/artifactory-cleanup.js
+ *
+ *
+ * The script uses your artifactory identity token
+ * which you should have saved in `.npmrc` (or `.yarnrc) file in your user home dir.
+ * The script reads the token using `yarn config get "//artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/:_authToken"` command
+ */
```
pbartkowiak-fandom, line 17: Personally, I'd leave that comment in the documentation instead of putting in here. At the end of the day, documentation is where we want to look for documentation.

scripts/artifactory-cleanup.js
```diff
@@ -0,0 +1,87 @@
...
+	return !packageName.match(/^\/jwplayer-fandom-\d+\.\d+\.\d+\.tgz/);
+}
+
+async function fetchTestPackages() {
+	const response = await fetch(`${BASE_ARTIFATORY_URL}${BASE_READ_PATH}`);
+	const json = await response.json();
+
+	const packages = json.children.map((child) => child.uri).filter(isTestVersion);
+
+	return packages;
```
pbartkowiak-fandom, line 61:
```suggestion
	return json.children.map((child) => child.uri).filter(isTestVersion);
```

README.md
```diff
@@ -164,6 +164,11 @@ Instead of using `/test-jw/src` app for testing the standalone player you can wo
 - Bump version and publish with `yarn pub:[patch|minor|major]` (e.g. `yarn pub:minor`).
 - If you don't have your artifactory credentials on `~/.npmrc` get them [here](https://fandom.atlassian.net/wiki/spaces/GEN/pages/110592255/Artifactory+-+Internal+package+repositories).
 
+## Removing unused test version
+
+- run `node scripts/artifactory-cleanup.js`
```
pbartkowiak-fandom, line 169: Please add the command to the package.json's `scripts` section.

nandy-andy (approved):

PR #195, opened 2023-09-12
COTECH-691 Introduce JwPlayerWrapperWithStrategyRules PoC for experiments

nandy-andy:
https://fandom.atlassian.net/browse/COTECH-691


src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -50,23 +57,35 @@ export const DesktopArticleVideoPlayerContent: React.FC<DesktopArticleVideoPlaye
...
+									<JwPlayerWrapper
+										getDismissed={getDismissed}
+										config={getArticleVideoConfig(videoDetails)}
+										onReady={(playerInstance) => {
+											articlePlayerOnReady(videoDetails, playerInstance);
+											setIsPlayerReady(true);
+										}}
+										stopAutoAdvanceOnExitViewport={false}
+									/>
+								)}
```
andrzejkupczyk, line 88: I'm not very familiar with React, but in Vue, we have dynamic components that simplify templates and reduce code duplication. Is there a way to achieve something similar in React? I did a quick search and found some common approaches like conditional rendering, using component maps, and utilizing functions or render props. In essence, is it feasible to reduce code duplication, or is the current implementation a temporary solution?

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,169 @@
...
+				onComplete();
+			}
+
+			// Incrementing videos watched count
+			videoIndexRef.current += 1;
+
+			// if stopAutoAdvanceOnExitViewport is false then we don't want to stop the auto advance, keep normal behavior
+			if (!stopAutoAdvanceOnExitViewport) {
+				return;
+			}
```
andrzejkupczyk, line 100: Is this code necessary? It seems like it does not do anything, as there is nothing after this condition anyway.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,169 @@
...
+		}
+	};
+
+	return (
+		<div className={className}>
+			<div
+				id={jwPlayerContainerEmbedId}
+				data-jw-placement-id={'21rL5wJF'}
+				data-jw-playlist={config.playlistUrl}
+				data-jw-recommendations_playlist_id={'FOhaD53w'}
```
andrzejkupczyk, line 160: I'm curious if it would be a good idea to place the placement ID and playlist ID in a configuration file or, at the very least, store them as constant variables.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,169 @@
...
+				data-jw-preroll_ad_tag={
+					'https://pubads.g.doubleclick.net/gampad/ads?iu=%2F5441%2Fwka1b.VIDEO%2Ffeatured%2Fdesktop%2Fucp_desktop-fandom-fv-article%2F_project43-life&sz=640x480&gdfp_req=1&output=xml_vast4&unviewed_position_start=1&env=vp&cust_params=src%3Dtest%26pos%3Dfeatured%26post_id%3D-1'
```
andrzejkupczyk, line 162: Just to be sure, is this tag correct?

src/jwplayer/utils/useAdEngineComplete.ts
```diff
@@ -5,14 +5,16 @@ import { race, timer } from 'rxjs';
 import { first } from 'rxjs/operators';
 import { recordVideoEvent, VIDEO_RECORD_EVENTS } from 'jwplayer/utils/videoTimingEvents';
 
-export default function useAdComplete(): boolean {
+export default function useAdEngineComplete(): boolean {
 	const [adComplete, setAdComplete] = useState(false);
 
 	useEffect(() => {
 		recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_AD_ENG_OPT_IN_LISTEN_START);
+		// TODO: we probably can remove this step as AdEngine won't send "[Ad Engine] Setup JWPlayer" before the consent
```
andrzejkupczyk, line 13: This comment seems more like a personal reminder, and it's not clear what action needs to be taken. If we have the opportunity to address this issue, why not go ahead and do so? If there is a related Jira ticket or GitHub issue, it would be helpful to include a link here.

src/utils/waitFor.ts
```diff
@@ -0,0 +1,42 @@
+interface waitForEventDescription {
+	eventCheck: () => boolean;
+	maxWaitTime?: number;
+	intervalTime?: number;
```
andrzejkupczyk, line 4: Is it possible to consider renaming these properties to include the time unit instead of using the 'Time' suffix? For instance, `maxWaitMilliseconds`?

src/jwplayer/utils/useJwpAdsSetupComplete.ts
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
+import { communicationService, ofType } from 'jwplayer/utils/communication';
+import { first } from 'rxjs/operators';
```
latata, line 3: I think you should rebase your branch from master as `rxjs` is no longer a thing here ;)

src/jwplayer/utils/useJwpAdsSetupComplete.ts
```diff
@@ -0,0 +1,23 @@
...
+import { recordVideoEvent, VIDEO_RECORD_EVENTS } from 'jwplayer/utils/videoTimingEvents';
+
+export default function useJwpAdsSetupComplete(): boolean {
+	const [jwpAdsSetupComplete, setJwpAdsSetupComplete] = useState(false);
```
latata, line 7: I noticed that it is common in this repo to have a hook with state and `useEffect` that waits for some events and sets the state value. IMO it's not a good approach, especially when the hooks are used multiple times - we unnecessarily create duplicated states and duplicated event handlers. I think we should use react context instead or introduce some kind of state management library. Not necessarily something to do now but we should refactor it at some point.

andrzejkupczyk: I've also observed that. Generally, the use of `useEffect` should be minimized since it's often unnecessary. However, I agree that discussing this falls beyond the scope of this PR, as you mentioned.

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -50,23 +57,35 @@ export const DesktopArticleVideoPlayerContent: React.FC<DesktopArticleVideoPlaye
...
+									/>
+								) : (
+									<JwPlayerWrapper
+										getDismissed={getDismissed}
+										config={getArticleVideoConfig(videoDetails)}
+										onReady={(playerInstance) => {
+											articlePlayerOnReady(videoDetails, playerInstance);
+											setIsPlayerReady(true);
+										}}
+										stopAutoAdvanceOnExitViewport={false}
```
pbartkowiak-fandom, line 86: Since the `onReady` function is the same on lines 72 and 82, you could write it once above and pass it as a prop to the components.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,169 @@
...
+			const script = document.createElement('script');
+			script.async = true;
+			script.src = 'https://cdn.jwplayer.com/v2/sites/cGlKNUnj/placements/embed.js';
+			script.onload = onload;
+			document.getElementsByTagName('head')[0].appendChild(script);
+		}
+	};
+
+	return (
+		<div className={className}>
```
pbartkowiak-fandom, line 155: This can result in `<div class=undefined'> after the render, which is not elegant.

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
nandy-andy: My hope is that it's just a temporary and after we conduct the experiment we'll remove `JwPlayerWrapper` component 😈 But if for some reason we'll learn that we need both options it's worth looking more into the solutions you've found 👌

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: Good question. I'll look into it at the end and remove if it's not necessary 👌

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
andrzejkupczyk: Sure, that sounds good. Thanks for the clarification.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: ACK. These will be dynamic in the future most likely 😊

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: That's the default 30s blue video. I'll replace it with what I get from [the AdEngine part](https://github.com/Wikia/ad-engine/pull/2141/files#diff-0da6adc9579963d840ecce8ccb2b4ca0a0b28efd4704bdcf64ec01804813091fR32) 😊

src/jwplayer/utils/useAdEngineComplete.ts
nandy-andy: Yeah, it's something like that. A reminder ;) We can't take actions now and we will after we conduct the experiment.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
andrzejkupczyk: Does this imply that the current PR is not ready for merging? Or are we planning to have a follow-up PR?

src/jwplayer/utils/useJwpAdsSetupComplete.ts
nandy-andy: Great feedback! I'll take a look at `PlayerContext` I saw in our code and see if this is something we could base on.

src/utils/waitFor.ts
nandy-andy: Let's see if after recent JWP changes we'll still need this interface and function 😈

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: Yes 😈 Hi hi hi

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: Can it? I barefacedly copied it from `JwPlayerWrapper` and I can see on my localhost it creates a `<div>` without class, so maybe something changed in React 🤔

yarn.lock
andrzejkupczyk, line 1: Are the changes in the lock file intentional? I don't see any corresponding changes in the `package.json` file 🤔

nandy-andy: Nah... maybe it got updated after merging `origin/master` and running `yarn install` (because in the master we removed rxjs this week).

test-jw/src/communication/ofType.js
```diff
@@ -0,0 +1,7 @@
+import { filter } from 'rxjs/operators';
+
+export function ofType(...types) {
```
latata, line 3: is this used anywhere?

test-jw/src/adEngineMock.js
```diff
@@ -0,0 +1,39 @@
+import { communicationService } from "./communication/communicationService";
```
latata, line 1: can't we use communication service from jwplayer-fandom package? 🤔

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,159 @@
...
+			}
+		});
+
+		playerInstance.on(JWEvents.COMPLETE, () => {
+			if (typeof onComplete === 'function') {
+				onComplete();
+			}
+
+			// Incrementing videos watched count
+			videoIndexRef.current += 1;
```
latata, line 97: is it used somewhere?

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,159 @@
...
+	shouldLoadSponsoredContentList = true,
+	jwPlayerContainerEmbedId = 'featured-video__player',
+	vastUrl,
+}) => {
+	const { setPlayer, setConfig } = useContext(PlayerContext);
+	const videoIndexRef = React.useRef(0);
+	const strategyRulesPlacementId = '21rL5wJF';
+	const recommendationPlaylistId = 'FOhaD53w';
+
+	useEffect(() => {
```
latata, line 34: It could be shared between both JwPlayerWrappers. Not sure as a part of this PR as this is only PoC 🤷‍♂️

nandy-andy: I tried and the new hook seems to work fine. Take a look please and let me know if there are no red flags 😊 https://github.com/Wikia/jwplayer-fandom/pull/195/commits/33dcedba3c88af4354192076058d515aafbdae32

test-jw/src/adEngineMock.js
nandy-andy: I'm not sure. It's not exported to the main package and this way we imitate more the actual apps that create a similar wrappers around https://github.com/Wikia/post-quecast 🤔 It's also too late for me to make it work 😅

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
latata: LGTM

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,128 @@
...
+				console.debug('Player ready...');
+				onReady(playerInstance);
+			}
+		});
+
+		playerInstance.on(JWEvents.COMPLETE, () => {
+			if (typeof onComplete === 'function') {
+				onComplete();
+			}
+		});
```
andrzejkupczyk, line 110: I'm wondering if it might be better, performance-wise, to avoid registering a callback if onComplete is not a function. In other words, perhaps we should do this conditionally?

```suggestion
		if (typeof onComplete === 'function') {
			playerInstance.on(JWEvents.COMPLETE, () => onComplete());
		}
```

src/jwplayer/types.d.ts
```diff
@@ -439,6 +450,7 @@ export interface JwPlayerWrapperProps extends JwPlayerContainerId {
 	className?: string;
 	stopAutoAdvanceOnExitViewport?: boolean;
 	shouldLoadSponsoredContentList?: boolean;
+	vastUrl?: boolean | string;
```
andrzejkupczyk, line 453: Why should this property be a boolean?

src/jwplayer/utils/useJwpWrapperInit.ts
```diff
@@ -0,0 +1,44 @@
...
+			};
+			retrieveSponsoredVideo().catch((e) => {
+				console.error('There was an issue with retrieving Sponsored Videos. ', e);
+			});
+		} else {
+			console.debug('Loading of Sponsored Content Video List was disabled.');
+		}
+		recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_INIT_RENDER);
+		loadEmbed();
+		setLoadEmbedExecuted(true);
```
andrzejkupczyk, line 40: I've just started learning React, so I apologize if my question seems... "basic" :D If the commented lines rely on the outcome of `retrieveSponsoredVideo`, will they function correctly? I recently read an [article about using async functions in watchEffect](https://devtrium.com/posts/async-functions-useeffect), where the author mentioned some important considerations.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: I think it makes sense, thanks :)

src/jwplayer/types.d.ts
nandy-andy: Good question. It should be string 🤔 I'll take a look 👍

src/jwplayer/utils/useJwpWrapperInit.ts
nandy-andy: I'm also not a React expert 😊 Thanks for sharing the article! I give it a go and my understanding is that since we don't use the results from the async call in the state we don't need to worry here 🤔 The code has changed but indeed it could've been a source of race-condition 👍

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,116 @@
...
+	};
+
+	const initPlayer = () => {
+		console.debug('Strategy rules enabled: the embed will be loaded inline', vastUrl);
+	};
+
+	const { setPlayer, setConfig } = useContext(PlayerContext);
+	useBeforeJwpWrapperRendered(initPlayer, shouldLoadSponsoredContentList);
+
+	const selector = `.${parentClassName} .strategyRulesWrapper`;
```
latata, line 95: why are we inserting the script tag in this div instead of `head`?

src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -178,11 +154,7 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 		}
 	};
 
-	return (
-		<div className={className}>
-			<div id={jwPlayerContainerEmbedId} />
-		</div>
-	);
+	return <div className={className}>{jwpWrapperInitialized && <div id={jwPlayerContainerEmbedId} />}</div>;
```
latata, line 157: why did you add `jwpWrapperInitialized` check here?

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,116 @@
...
+	className,
+	shouldLoadSponsoredContentList = true,
+	jwPlayerContainerEmbedId = 'featured-video__player',
+	vastUrl,
+	parentClassName,
+}) => {
+	const strategyRulesPlacementId = '21rL5wJF';
+	const recommendationPlaylistId = 'FOhaD53w';
+
+	const registerEventHandlers = (playerInstance) => {
```
latata, line 35: can't have it shared between with and without strategy rules? 🤔

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,116 @@
...
+		window.jwplacements._getPlacementReadyPromise(strategyRulesPlacementId).then(jwPlayerLoaded);
+	};
+
+	useScript(strategyRulesUrl, selector, onBeforeLoad, {
+		onLoad,
+		className,
+		id: jwPlayerContainerEmbedId,
+	});
+
+	return <></>;
```
latata, line 113: I think it can just return `null`

src/jwplayer/utils/useAdEngineComplete.ts
```diff
@@ -3,15 +3,17 @@ import { useState, useEffect } from 'react';
...
 	const [adComplete, setAdComplete] = useState(false);
 	const communicationService = getCommunicationService();
 
 	useEffect(() => {
 		recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_AD_ENG_OPT_IN_LISTEN_START);
+		// TODO: we probably can remove this step as AdEngine won't send "[Ad Engine] Setup JWPlayer" before the consent
 		communicationService.on('[AdEngine OptIn] set opt in', () => {
 			recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_AD_ENG_OPT_IN_MESSAGE_RECIEVED);
 			recordVideoEvent(VIDEO_RECORD_EVENTS.JW_PLAYER_AD_ENG_CONFIG_LISTEN_START);
+			// TODO: we probably can remove this after we finish testing strategy rules
```
dzawada, line 16: To which lines does this comment refer to?

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: I can try adding it to the `head` but I'm afraid it won't work as I imagine it'll insert JWP divs in the head or throw an error 🤔  

This is the "inline method" of integrating strategy rules. The "header method" I tried before was injecting the `embed.js` script in the `head`. But it didn't work that well because it scanned the DOM only once on page load and required divs with placements attributes to be already on the page. We were experiencing a race-condition.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: I think I've been thinking about it and decided not to do it now. We're not sure if the strategy rules are the right solution and if they are, they'll replace the previous one, so maybe it's too early to share the code 🤔

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: I can try but I think I had it like that before and Typescript was complaining ;)

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -0,0 +1,116 @@
...
+	getDismissed = () => false,
+	onComplete,
+	className,
+	shouldLoadSponsoredContentList = true,
+	jwPlayerContainerEmbedId = 'featured-video__player',
+	vastUrl,
+	parentClassName,
+}) => {
+	const strategyRulesPlacementId = '21rL5wJF';
+	const recommendationPlaylistId = 'FOhaD53w';
```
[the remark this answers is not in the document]
nandy-andy: AFAIK it'll become dynamic in the future. I think the idea is that once we check the strategy rules solution we'll start passing more custom parameters to it (for example vertical) on which the strategy will pick the right recommendation playlist and we'll remove this line.

src/jwplayer/players/shared/JwPlayerWrapper.tsx
nandy-andy: I thought we didn't want to render the `<div />` before we have signal from the hook JWP has been initialised. But since the flag only means that the `initPlayer()` was called (it doesn't mean the JWP script has been loaded or the sponsored videos has been retrieved) I should change it and just execute the `useBeforeJwpWrapperRendered()` hook 🤔 Thanks for catching! :)

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -50,23 +58,31 @@ export const DesktopArticleVideoPlayerContent: React.FC<DesktopArticleVideoPlaye
...
-								<JwPlayerWrapper
-									getDismissed={getDismissed}
-									config={getArticleVideoConfig(videoDetails)}
-									onReady={(playerInstance) => {
-										articlePlayerOnReady(videoDetails, playerInstance);
-										setIsPlayerReady(true);
-									}}
-									stopAutoAdvanceOnExitViewport={false}
-								/>
+								{jwpAdsSetupComplete.strategyRulesEnabled && <div className="strategyRulesWrapper" />}
```
latata, line 69: IMO it should be somewhere inside `<JwPlayerWrapperWithStrategyRules>` component or we should create another wrapping component that will contain this div and `<JwPlayerWrapperWithStrategyRules>`

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
latata, line 95: We should rather use react's ref to pass the element reference instead of using class selectors.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
nandy-andy: Hm, I think it's OK. Thanks!

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -50,25 +58,43 @@ export const DesktopArticleVideoPlayerContent: React.FC<DesktopArticleVideoPlaye
...
+							adEngineComplete && (
+								<div>
+									<div className={styles.topBar}>
+										{!isScrollPlayer && <UnmuteButton />}
+										{isScrollPlayer && (
+											<CloseButton
+												dismiss={() => setDismissed(true)}
+												iconColor={'#fff'}
+												className={styles.closeButton}
+											/>
```
latata, line 86: do we need to have this code duplicated? 🤔

latata (approved):

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
nandy-andy: 👍 I even started writing a `DesktopArticleVideoFrame` component ;) But then again once we decide if we want to go with strategy rules globally or not it all will be unnecessary 😊

dzawada (approved):

PR #196, opened 2023-09-13
COTECH-136 (step2) | configurable mobile scroll position

latata:
* Added new prop for `MobileArticleVideoLoader`: `scrollTopPosition` - a string that is passed as `top` position for the container in scroll mode
* Passed through MobileArticleVideoContext to avoid unnecessary prop passing down 
* Removed unused props `hasPartnerSlot` and `isFullScreen` and commented out tracking code

### DIFF
--- a/src/loaders/types.d.ts
+++ b/src/loaders/types.d.ts
@@ -5,7 +5,11 @@ import {
 } from 'jwplayer/types';
 
 export type DesktopArticleVideoLoaderProps = DesktopArticleVideoPlayerProps | undefined;
-export type MobileArticleVideoLoaderProps = MobileArticleVideoPlayerProps | undefined;
+export type MobileArticleVideoLoaderProps =
+	| (MobileArticleVideoPlayerProps & {
+			scrollTopPosition?: string;
+	  })
+	| undefined;
 export type CanonicalVideoLoaderProps = CanonicalVideoPlayerProps | undefined;
 
 interface MWConfig {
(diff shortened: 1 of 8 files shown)


nandy-andy (approved):

src/loaders/types.d.ts
```diff
@@ -5,7 +5,11 @@ import {
 } from 'jwplayer/types';
 
 export type DesktopArticleVideoLoaderProps = DesktopArticleVideoPlayerProps | undefined;
-export type MobileArticleVideoLoaderProps = MobileArticleVideoPlayerProps | undefined;
+export type MobileArticleVideoLoaderProps =
+	| (MobileArticleVideoPlayerProps & {
+			scrollTopPosition?: string;
+	  })
+	| undefined;
```
andrzejkupczyk, line 12: It's a minor issue, but could we improve readability by removing the extra pipe character and adjusting the indentation?

src/youtube/players/YoutubeMobileArticleVideoPlayer.tsx
```diff
@@ -34,7 +32,7 @@ const MobileArticleVideoWrapper: React.FC<MobileArticleVideoWrapperProps> = ({
 						: styles.mobileArticleVideoWrapperIsNotScrollPlayer,
 				)}
 				style={{
-					...(isScrollPlayer && { top: topPosition }),
+					...(isScrollPlayer && { top: scrollTopPosition }),
```
andrzejkupczyk, line 35: Considering the type, it seems that `scrollTopPosition` could potentially become `undefined` in both this context and in `src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx`. Do you think we should address this situation?

latata: The value in the context should always be set but even if it is undefined it will just not have `top` css property. I think it's fine

src/loaders/types.d.ts
latata: I'm pretty sure it was generated by prettier - so I would not touch it.

andrzejkupczyk: I believe you're pretty right. Thanks for pointing that out!

PR #198, opened 2023-09-13
When switching tabs while an ad is playing, do *not* pause the ad playback

pbartkowiak-fandom:
https://fandom.atlassian.net/browse/COTECH-654

----------

Related to https://github.com/Wikia/ad-engine/pull/2150 PR.

We want to control the behavior of ads from JWPlayer. In this case, we want the ad to continue playing when the tab is changed.

### DIFF
--- a/src/jwplayer/players/shared/JwPlayerWrapper.tsx
+++ b/src/jwplayer/players/shared/JwPlayerWrapper.tsx
@@ -112,8 +112,11 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 
 			playerInstance.on(JWEvents.PLAY, ({ playReason, viewable }: JWPlayEvent) => {
 				const dismissed = getDismissed();
-				// Pause the content play when the user closed the mini player playing the ad
 				if (dismissed && viewable === 0 && playReason === 'autostart') {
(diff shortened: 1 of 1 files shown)


src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -104,13 +104,19 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 				// Keep playing the ad when the user closed the mini player
 				if (viewable === 0 && pauseReason === 'external') {
 					playerInstance.play();
+				} else if (document.visibilityState === 'hidden' && pauseReason === undefined) {
+					// Keep the ad playing even when the user switches tabs
+					playerInstance.play();
```
latata, line 109: shouldn't we just remove this thing instead of making these changes? https://github.com/Wikia/ad-engine/blob/8712df48ac604fd89e887e9d4747b9acc201b72d/src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts#L147C50-L147C73

pbartkowiak-fandom: Nice finding. Yes, I think it's worth trying 👍

PR #199, opened 2023-09-14
COTECH-716 bugfix

David-Iacovone:


agatakozinska (approved):

src/jwplayer/players/MobileArticleVideoPlayer/Attribution.tsx
```diff
@@ -14,7 +14,7 @@ const Attribution: React.FC = () => {
 			<a className={styles.usernameLink} href={playlistListItem.userUrl}>
 				{playlistListItem.username}
 			</a>
-			<a className={styles.iconLink} href={playlistListItem.userUrl}>
+			<a aria-label={'Attribution Link'} className={styles.iconLink} href={playlistListItem.userUrl}>
```
latata, line 17: I'd set `aria-label` to `playlistListItem.username` or if we want to have something more descriptive we should apply i18n

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@fandom/jwplayer-fandom",
-	"version": "2.7.0",
```
latata, line 3: I think we should not publish new version without ticket number at the end until the ticket is tested. We should publish just before merging to master

src/jwplayer/players/DesktopArticleVideoPlayer/Attribution.module.css
```diff
@@ -4,6 +4,7 @@
 	display: flex;
 	justify-content: center;
 	padding: 6px;
+	cursor: pointer;
```
latata, line 7: we should not add `cursor: pointer` wrapper because it's not clickable

David-Iacovone: I added it here to avoid style duplication

package.json
David-Iacovone: Alright, we can do that

src/jwplayer/players/MobileArticleVideoPlayer/Attribution.tsx
David-Iacovone: Yeah that's a good idea

src/jwplayer/players/DesktopArticleVideoPlayer/Attribution.module.css
David-Iacovone: I can change that. No prob

latata (commented):
Why are there changes in yarn.lock files? 🤔

David-Iacovone:
> Why are there changes in yarn.lock files? 🤔

I see that some babel deps changed. I just did yarn install so something probably changed in babel while I was coding

src/jwplayer/players/DesktopArticleVideoPlayer/Attribution.module.css
```diff
@@ -8,10 +8,12 @@
 
 .usernameLink {
 	color: rgba(var(--theme-page-text-color--rgb), 0.75);
+	cursor: pointer;
 }
 
 .iconLink {
 	color: rgba(var(--theme-page-text-color--rgb), 0.75);
+	cursor: pointer;
```
latata, line 16: what's the point of adding `cursor: pointer` here? These are `<a>` elements with `href` which by default have pointer cursor.

David-Iacovone: Yeah, we can remove it. I was not aware that they have this by default

latata (approved):

PR #207, opened 2023-10-06
TSKFRC-109 Optimise the way JW Player checks for video takeovers

David-Iacovone:
**Ticket:** https://fandom.atlassian.net/browse/TSKFRC-109

**Remark 1:** Due to the changes here it will no longer be possible to call for a forced takeover on any article page. However, we still have test article pages with enabled vimeo and youtube takeovers. Therefore, I think this is not an issue.

**Remark 2:** Apart from changes requested in the ticket I added a new script in package.json `yarn dev` It is a combination of `yarn build` and `yarn app`.


package.json
```diff
@@ -18,6 +18,7 @@
 		"deploy-stand-alone": "ts-node --project stand-alone/tsconfig-stand-alone.json stand-alone/deploy/deploy.ts",
 		"watch": "yarn run create-dist && NODE_ENV=development rollup -c --watch",
 		"app": "cd test-jw && yarn && yarn start",
+		"dev": "yarn build && yarn app",
```
latata, line 21: I've added `start` script that do the same thing on my branch ;)

David-Iacovone: Nice, we need to choose which one to keep then.

David-Iacovone: I actually avoided watch since @nandy-andy said we have experienced issues with it.

src/loaders/utils/GetTakeoverDetails.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export async function getTakeoverDetails({ deviceType }: YoutubePlayerTrackingProps) {
+	const { wikiId, isTier3Wiki } = getMediaWikiConfigDetails();
+	if (!wikiId || isTier3Wiki) {
+		return null;
+	}
+
+	const response = await fetch(getTakeoverUrl(wikiId));
+	const dataArray = await response.json();
+	if (dataArray.length === 0) {
```
latata, line 27:
```const [data] = await response.json();
if (!data) {
```

src/loaders/utils/GetTakeoverDetails.ts
```diff
@@ -0,0 +1,96 @@
...
+		return null;
+	}
+
+	if (dataArray[0].youtube_take_over) {
+		const youtubeDetails: YoutubeTakeOverDetails = { isYoutubeTakeover: false };
```
latata, line 32: Can we clean up the typing a bit? I think we don't need separate types for YouTube and Vimeo as both have videoId and type flag. I think we could have
```
interface TakeoverDetails {
  videoId: string;
  type: 'youtube' | 'vimeo';
}

Comment (latata) [reply src/loaders/utils/GetTakeoverDetails.ts:32]: this way we could avoid a lot of code duplication here

**Review by latata — CHANGES_REQUESTED** [review]

File: package.json (hunk 1 above)
Comment (latata) [reply package.json:21]: Watch works fine for me. I have never noticed any issues

File: src/loaders/utils/GetTakeoverDetails.ts [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export async function getTakeoverDetails({ deviceType }: YoutubePlayerTrackingProps) {
+	const { wikiId, isTier3Wiki } = getMediaWikiConfigDetails();
+	if (!wikiId || isTier3Wiki) {
+		return null;
+	}
+
+	const response = await fetch(getTakeoverUrl(wikiId));
+	const dataArray = await response.json();
```
latata, line 26: please add a type of the response

src/loaders/utils/GetTakeoverDetails.ts
David-Iacovone: But will this guard from an empty array?

src/loaders/utils/GetTakeoverDetails.ts
David-Iacovone: Good idea, let me clean this up

src/loaders/utils/GetTakeoverDetails.ts
David-Iacovone: sure

src/loaders/utils/GetTakeoverDetails.ts
latata: also, I think these `eligibleFor*Takeover` are redundant and we can just have a check for videoId length in `getTakeoverDetails`. Then in the loaders code we could just check if `takeoverDetails?.type === 'vimeo'` (or youtube)

package.json
nandy-andy: I think the old readme mentioned issues but let's experience them by ourselves and then think about a fix ;)

src/loaders/utils/GetTakeoverDetails.ts
```diff
@@ -0,0 +1,57 @@
...
+	const { wikiId, isTier3Wiki } = getMediaWikiConfigDetails();
+	if (!wikiId || isTier3Wiki) {
+		return null;
+	}
+
+	const response: Response = await fetch(getTakeoverUrl(wikiId));
+	const dataArray: TakeoverResponse[] = await response.json();
+	const takeoverDetails: TakeoverDetails = { videoId: null, type: null };
+
+	if (dataArray.length === 0 || !dataArray || isServerSide()) {
```
latata, line 28: Do you know what's the purpose of this `isServerSide()` check? I think we actually never use the player on the server side. 

Also I noticed that this function is actually broken:
```const isServerSide = () => {
	return typeof window === undefined;
};```
undefined is not wrapped with `'` so it will always return false. IMO we can remove this function and all of the usages. :)

File: src/loaders/utils/GetTakeoverDetails.ts [hunk 6]
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export async function getTakeoverDetails({ deviceType }: YoutubePlayerTrackingProps) {
+	const { wikiId, isTier3Wiki } = getMediaWikiConfigDetails();
+	if (!wikiId || isTier3Wiki) {
+		return null;
+	}
+
+	const response: Response = await fetch(getTakeoverUrl(wikiId));
+	const dataArray: TakeoverResponse[] = await response.json();
```
latata, line 25: const [data] = await response.json() as TakeoverResponse[];

src/loaders/utils/GetTakeoverDetails.ts
```diff
@@ -0,0 +1,57 @@
...
+		takeoverDetails.videoId = dataArray[0].takeover_video_id;
+	}
+	if (dataArray[0].vimeo_take_over) {
+		takeoverDetails.type = 'vimeo';
+		takeoverDetails.videoId = dataArray[0].takeover_video_id;
+	}
+
+	switch (takeoverDetails.type) {
+		case 'youtube': {
+			console.debug('Youtube Takeover: Eligible for youtube takeover based on the targeting params.', takeoverDetails);
```
latata, line 43: TBH I don't see the reason to keep these console.debugs 🤷‍♂️

src/loaders/utils/GetTakeoverDetails.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export async function getTakeoverDetails({ deviceType }: YoutubePlayerTrackingProps) {
+	const { wikiId, isTier3Wiki } = getMediaWikiConfigDetails();
+	if (!wikiId || isTier3Wiki) {
+		return null;
+	}
+
+	const response: Response = await fetch(getTakeoverUrl(wikiId));
+	const dataArray: TakeoverResponse[] = await response.json();
+	const takeoverDetails: TakeoverDetails = { videoId: null, type: null };
```
latata, line 26: I think it's better to return undefined or null if there is no takeover.

So we could have:

```
if(!data) {
  return;
}

return {
  type: data.youtube_take_over ? 'youtube' : 'vimeo',
  videoId: data.takeover_video_id
}

File: src/loaders/utils/GetTakeoverDetails.ts [hunk 9]
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if (dataArray[0].vimeo_take_over) {
+		takeoverDetails.type = 'vimeo';
+		takeoverDetails.videoId = dataArray[0].takeover_video_id;
+	}
+
+	switch (takeoverDetails.type) {
+		case 'youtube': {
+			console.debug('Youtube Takeover: Eligible for youtube takeover based on the targeting params.', takeoverDetails);
+			trackYoutubeTakeoverDetails({ deviceType, videoId: takeoverDetails.videoId });
```
latata, line 44: Do you know why do we track youtube takeover and not the vimeo takeover? 🤔

David-Iacovone: not sure, however I did not want change the existing behaviour. Perhaps it's needed for ad-engine

David-Iacovone: Well, at second thought there useful but if we can remove `trackYoutubeTakeoverDetails()` then I think we can remove this switch entirely. So let's give it a shot. Unless, @nandy-andy knows something about it.

src/loaders/utils/GetTakeoverDetails.ts
David-Iacovone: Done

src/loaders/utils/GetTakeoverDetails.ts
David-Iacovone: As mentioned below, I removed the whole debug statement

src/loaders/utils/GetYoutubeTakeoverDetails.ts
```diff
@@ -1,68 +0,0 @@
...
-		wikiYoutubeTakeoverDetails?.takeover_video_id?.trim().length !== 0
-	) {
-		youtubeTakeoverDetails.isYoutubeTakeover = wikiYoutubeTakeoverDetails.youtube_take_over;
-		youtubeTakeoverDetails.youtubeVideoId = wikiYoutubeTakeoverDetails.takeover_video_id;
-		console.debug(
-			'Youtube Takeover: Eligible for youtube takeover based on the targeting params.',
-			youtubeTakeoverDetails,
-			wikiYoutubeTakeoverDetails,
-		);
-		trackYoutubeTakeoverDetails({ deviceType, youtubeVideoId: youtubeTakeoverDetails.youtubeVideoId });
```
latata, line 57: is this tracking no longer needed? 🤔

David-Iacovone: I do not think is needed, especially considering that this is the only reason to keep this switch case. Unless @nandy-andy knows more about it.

David-Iacovone: Alright, I will ask @jasesanders on daily.

latata (approved):

PR #209, opened 2023-10-09
Use Vite instead of create-react-app

latata:


package.json
```diff
@@ -18,12 +18,12 @@
 	"types": "./dist/main.d.ts",
 	"install-peers": "install-peers",
 	"scripts": {
-		"start": "npm-run-all --parallel watch app",
+		"start": "yarn build-fandom-player && npm-run-all --parallel watch app",
```
latata, line 21: run build before starting the app so the lib is ready when the app is starting

package.json
```diff
@@ -18,12 +18,12 @@
...
 		"build": "yarn build-fandom-player && yarn build-stand-alone",
 		"build-fandom-player": "yarn run create-dist && rollup -c",
 		"build-stand-alone": "cd stand-alone && yarn && yarn run create-standalone-dist && rollup -c rollup-stand-alone.config.js",
 		"deploy-stand-alone": "ts-node --project stand-alone/tsconfig-stand-alone.json stand-alone/deploy/deploy.ts",
-		"watch": "yarn run create-dist && NODE_ENV=development rollup -c --watch",
+		"watch": "NODE_ENV=development rollup -c --watch",
```
latata, line 26: do not remove the directory on watch to make the test app work immediately after starting

test-jw/package.json
```diff
@@ -1,39 +1,20 @@
 {
-  "name": "my-app",
+  "name": "jwplayer-fandom-test-app",
   "version": "0.1.0",
```
andrzejkupczyk, line 3: Perhaps it's a good opportunity to get rid of this property? I believe it's optional, but if it's left, it would need to be updated.

andrzejkupczyk (approved):

dzawada (approved):

PR #214, opened 2023-11-22
COTECH-832 Pass VAST XML to the advertisement config when available

nandy-andy:
These changes are strongly connected with https://github.com/Wikia/ad-engine/pull/2252. We don't want JWPlayer to repeat ad server request once we got it from the "scout request" introduced in the linked PR.


src/jwplayer/players/shared/JwPlayerWrapper.tsx
```diff
@@ -36,6 +36,7 @@ const JwPlayerWrapper: React.FC<JwPlayerWrapperProps> = ({
 	stopAutoAdvanceOnExitViewport,
 	shouldLoadSponsoredContentList = true,
 	jwPlayerContainerEmbedId = 'featured-video__player',
+	vastXml = '',
```
latata, line 39: IMO we don't need a default empty string value. Can we just have it undefined as default?

src/jwplayer/utils/useJwpAdsSetupComplete.ts
```diff
@@ -21,17 +23,19 @@ export default function useJwpAdsSetupComplete(): JwpAdsSetupCompleteResult {
 	const [jwpAdsSetupComplete, setJwpAdsSetupComplete] = useState(null);
 	const [vastUrl, setVastUrl] = useState(null);
 	const [strategyRulesEnabled, setStrategyRulesEnabled] = useState(false);
+	const [vastXml, setVastXml] = useState('');
```
latata, line 26: Same here. IMO vastXml could be undefined by default.
`const [vastXml, setVastXml] = useState<string>();`

dzawada (approved):

PR #215, opened 2023-11-23
COTECH-863 | Clean up tracking, mark tracking-metrics lib as a peer dependency

latata:
https://fandom.atlassian.net/browse/COTECH-863


dzawada (approved):

src/jwplayer/utils/videoTracking.ts
```diff
@@ -79,7 +78,7 @@ export const PROPERTY_NAMES = {
 };
 
 function addRunTimeParams(trackingParams: TrackData): TrackData {
-	const trackDataObject: TrackData = { ...addGlobalProps(), ...trackingParams };
```
latata, line 82: I don't know why these global props were added here but they are basically duplicated from tracking lib so I removed it from here.

andrzejkupczyk: I appreciate the clarification. However, for future reference, it might be advisable to include a `@todo` comment instead and avoid altering the code within unrelated PR since, as you mentioned, the reason behind adding these global properties is unknown 😉

andrzejkupczyk (approved):

nandy-andy (approved):

src/jwplayer/utils/videoTracking.ts
latata: Actually, it's a bit related because these global props overwrite user id as well but instead of removing just this one I removed all of them. :) 

Also, I like to follow "The Boy Scouting Rule" (https://levelup.gitconnected.com/the-boy-scouting-rule-a-simple-yet-effective-programming-principle-5d504c7dd993) so I always try to clean up instead of putting todo comments if possible

nandy-andy: Thanks for clarification! It helps more than the original PR when it was added ;) https://github.com/Wikia/jwplayer-fandom/pull/85

andrzejkupczyk: I'm aware of the rule, but due to its P1 status, I strongly recommend we exercise additional caution in the future 😄

PR #216, opened 2023-12-04
COTECH-581 | Use hasAds in standalone player

latata:
https://fandom.atlassian.net/browse/COTECH-581

### DIFF
--- a/src/jwplayer/types.d.ts
+++ b/src/jwplayer/types.d.ts
@@ -486,6 +486,7 @@ export interface RedVentureVideoPlayerProps extends JwPlayerContainerId {
 	showScrollPlayer: boolean;
 	playerUrl?: string;
 	autoPlay?: boolean;
+	hasAds?: boolean;
 }
 
 export interface MobileArticleVideoPlayerProps {
--- a/stand-alone/standalone-loader.tsx
+++ b/stand-alone/standalone-loader.tsx
@@ -8,9 +8,9 @@ import {
 	canPlayerRender,
 	getVideoDetails,
 	getVideoWrapperElement,
-	RedVenturePlayerContextProps
+	RedVenturePlayerContextProps,
 } from './loaderHelper';
-import { RedVentureVideoDetails, RedVentureVideoPlayerProps } from "./types";
+import { RedVentureVideoDetails, RedVentureVideoPlayerProps } from './types';
 
 console.debug('Standalone RedVenture Video Player ...');
 
@@ -21,7 +21,7 @@ interface WindowWithRedVentureJWPlayer extends Window {
 			playerName: string;
 		};
 		[key: string]: unknown;
-	},
+	};
 	loadPlayer: (props: RedVenturePlayerContextProps) => void;
 }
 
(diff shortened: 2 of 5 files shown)


stand-alone/standalone-loader.tsx
```diff
@@ -80,6 +80,7 @@ window.loadPlayer = async (context: RedVenturePlayerContextProps) => {
 			jwPlayerContainerEmbedId: getJwPlayerContainerEmbedId(context),
 			playerUrl: context?.playerUrl,
 			autoPlay: context?.autoPlay,
+			hasAds: context.hasAds,
```
andrzejkupczyk, line 83: Based on `RedVenturePlayerContext` interface, this parameter is optional and could potentially be undefined.

```suggestion
			hasAds: context?.hasAds,
```

src/jwplayer/types.d.ts
```diff
@@ -486,6 +486,7 @@ export interface RedVentureVideoPlayerProps extends JwPlayerContainerId {
 	showScrollPlayer: boolean;
 	playerUrl?: string;
 	autoPlay?: boolean;
+	hasAds?: boolean;
```
andrzejkupczyk, line 489: Should we also update `stand-alone/types.d.ts`?

stand-alone/standalone-loader.tsx
```diff
@@ -80,6 +80,7 @@ window.loadPlayer = async (context: RedVenturePlayerContextProps) => {
...
 		} as RedVentureVideoPlayerProps),
```
andrzejkupczyk, line 84: It seems that the conversion of type `{ videoDetails: RedVentureVideoDetails, ... }` to type `RedVentureVideoPlayerProps` could be a mistake, as these types are not compatible. I suspect that this is intentional. However, should we convert the expression to `unknown` first?

```ts
videoDetails: redVentureVideoDetails as unknown,
```

Currently, it should throw the TS2352 error 🤔

stand-alone/standalone-loader.tsx
latata: The `hasAds` is optional but `context` is not. I think these `?` above are redundant.

stand-alone/standalone-loader.tsx
latata: As that issue was already there I'd not bother as it's not critical (no runtime issue just types) and we are most likely sunsetting jwplayer-fandom soon :)

latata: ☝️ that is actually not true -adding `hasAds?: boolean;` to `stand-alone/types.d.ts` resolves the issue. I will add it

andrzejkupczyk: The error TS2352 has been resolved. Thank you very much!

src/jwplayer/types.d.ts
andrzejkupczyk: This is no longer applicable 👍🏻

stand-alone/standalone-loader.tsx
andrzejkupczyk: Thanks for the clarification. You could be right, sir. However, something still seems a bit off to me 🤔 I might have misunderstood the changes, but shouldn't its value default to true? If it's undefined, that wouldn't happen, would it? 😬

andrzejkupczyk: Discussed directly 👍🏻

PR #217, opened 2024-01-15
COTECH-816

David-Iacovone:


src/jwplayer/utils/iasTracker.ts
```diff
@@ -0,0 +1,25 @@
+import { scriptLoader } from 'jwplayer/utils/scriptLoader';
+
+const scriptUrl = '//static.adsafeprotected.com/vans-adapter-google-ima.js';
+class IasTracker {
+	private scriptPromise: Promise<Event>;
```
andrzejkupczyk, line 5: I'd suggest adding the `undefined` type to this property since it doesn't have an initializer and isn't assigned within the constructor.

src/jwplayer/utils/iasTracker.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	public async loadIasTrackerIfEnabled(skin: 'fandomdesktop' | 'fandommobile'): Promise<void> {
+		const deserializedConfigData = JSON.parse(localStorage.getItem(`instant-config-${skin}`)).data;
```
andrzejkupczyk, line 8: `Storage.getItem()` returns a key's value or `null` if the key doesn't exist. Just a heads up, `JSON.parse` expects `string`, so it might be a good idea to add a check or handle this case separately.

src/jwplayer/utils/scriptLoader.ts
```diff
@@ -0,0 +1,54 @@
+class ScriptLoader {
+	createScript(
+		src: string,
+		type = 'text/javascript',
+		isAsync = true,
+		node: HTMLElement | string = null,
```
andrzejkupczyk, line 6: `null` can't be assigned to the type `string | HTMLElement`. Since you're always passing a string ("first") based on usage, we could just remove the default value, I think.

```suggestion
		node: HTMLElement | string,
```

src/jwplayer/utils/scriptLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		return script;
+	}
+
+	loadScript(
+		src: string,
+		type = 'text/javascript',
+		isAsync = true,
+		node: HTMLElement | string = null,
```
andrzejkupczyk, line 41: Same as above, `null` can't be assigned to the type `string | HTMLElement`

src/jwplayer/utils/scriptLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+		});
+
+		Object.keys(datasets).forEach((dataset) => {
+			script.dataset[dataset] = datasets[dataset];
+		});
+
+		if (typeof node === 'string') {
+			const temp: ChildNode = document.getElementsByTagName('script')[0];
+
+			temp.parentNode.insertBefore(script, temp);
```
andrzejkupczyk, line 27: Let's keep in mind that `temp.parentNode` might be `null` 😉

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
```diff
@@ -62,6 +66,9 @@ export const DesktopArticleVideoPlayerContent: React.FC<DesktopArticleVideoPlaye
 
 	const onPlayerInstanceReady = (playerInstance) => {
 		articlePlayerOnReady(videoDetails, playerInstance);
+		communicationService.on('[JWPlayer] Player Ready', async () => {
```
andrzejkupczyk, line 69: Is "player instance ready" a different stage compared to "player ready"? In other words, is there a specific reason to listen for this event within the "onPlayerInstanceReady" callback? 🤔

src/jwplayer/players/MobileArticleVideoPlayer/MobileArticleVideoPlayer.tsx
```diff
@@ -61,6 +65,9 @@ export const MobileArticleVideoPlayerContent: React.FC<MobileArticleVideoPlayerP
 
 	const onPlayerInstanceReady = (playerInstance) => {
 		articlePlayerOnReady(videoDetails, playerInstance);
+		communicationService.on('[JWPlayer] Player Ready', async () => {
```
andrzejkupczyk, line 68: Same as above, based on the callback's name, it's being called when the player instance is ready 🤔

David-Iacovone: Yes

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
David-Iacovone: It seems to me to be a good moment to call it. We can change it if you have other ideas.

src/jwplayer/utils/iasTracker.ts
```diff
@@ -0,0 +1,25 @@
+import { scriptLoader } from 'jwplayer/utils/scriptLoader';
+
+const scriptUrl = '//static.adsafeprotected.com/vans-adapter-google-ima.js';
+class IasTracker {
+	private scriptPromise: Promise<Event>;
+
+	public async loadIasTrackerIfEnabled(skin: 'fandomdesktop' | 'fandommobile'): Promise<void> {
```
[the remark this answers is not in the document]
David-Iacovone: You mean to create a separate util to parse the config?

src/jwplayer/utils/scriptLoader.ts
David-Iacovone: yea, will fix it. good point

src/jwplayer/players/DesktopArticleVideoPlayer/DesktopArticleVideoPlayer.tsx
andrzejkupczyk: The moment is perfect, but is there a specific reason we need to involve the communication service in this case? Why even bother if the player instance is ready?

src/jwplayer/utils/iasTracker.ts
David-Iacovone: sure

src/jwplayer/utils/iasTracker.ts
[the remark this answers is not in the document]
andrzejkupczyk: I meant that the tracker should probably remain unaware of the FCP, its various skins, or the user's device. This is important since our players are also used on other platforms and sites, right?

After some consideration, I suggest using the full config item key instead of a skin name. It would make it clearer if the IAS Tracker's status is based on the instant config key rather than on a specific 'skin'. What are your thoughts on this? Here's a quick suggestion for how it might look:

```suggestion
	public async loadIasTrackerIfEnabled(instantConfigKey: string): Promise<void> {
```

PR #221, opened 2024-02-09
COTECH-1069 Quick fix for ExCo player via JWP

nandy-andy:


src/jwplayer/types.d.ts
```diff
@@ -448,7 +448,7 @@ export interface JwPlayerWrapperProps extends JwPlayerContainerId {
 	getDismissed?: () => boolean;
 	config?: PlayerConfig;
 	playerUrl?: string;
-	onReady?: (playerInstance: Player) => void;
+	onReady?: (playerInstance?: Player) => void; // playerInstance for onReady is not needed in the SR scenario it is needed when AdEngine controls the ads
```
latata, line 451: how about prevent calling `articlePlayerOnReady` when playerInstance is undefined? 🤔

nandy-andy: do you mean preventing calling it but also moving the showing placeholder logic out of it? it may be a better long-term solution 👍

dzawada (approved):

PR #226, opened 2024-03-04
COTECH-1080 | rv param

latata:
https://fandom.atlassian.net/browse/COTECH-1080


src/jwplayer/utils/updateRVParam.ts
andrzejkupczyk, line 1: Why not use the URL API?

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -41,6 +42,8 @@ const JwPlayerWrapperWithStrategyRules: React.FC<JwPlayerWrapperProps> = ({
 	const strategyRulesPlacementId = 'KmMLkvao';
 	const recommendationPlaylistId = 'FOhaD53w';
 	const communicationService = getCommunicationService();
+	const prerollAdTag = vastUrl ?? '';
```
andrzejkupczyk, line 45: Using multiple names for the same thing can be confusing. We already have "vastUrl" and "adTagUrl" as different names for the VAST ad tag URL value. Let's avoid making things even more complicated and at least always include the "url" part 😉 

```suggestion
	const prerollAdTagUrl = vastUrl ?? '';
```

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
```diff
@@ -71,6 +74,16 @@ const JwPlayerWrapperWithStrategyRules: React.FC<JwPlayerWrapperProps> = ({
 			addBaseTrackingEvents(playerInstance);
 		});
 
+		playerInstance.on(JWEvents.AD_IMPRESSION, () => {
+			const newAdIndex = adIndex + 1;
+			const newPrerollAdTag = updateRVParam(prerollAdTag, newAdIndex);
```
andrzejkupczyk, line 79: Analogous request as the one in the previous comment. Terms "tag" and "url" represent different concepts.

src/jwplayer/players/shared/JwPlayerWrapperWithStrategyRules.tsx
latata: `preroll_ad_tag` is the name of JWP custom parameter (`jwDataStore.custom[strategyRulesPlacementId].preroll_ad_tag`) so I'd leave it as is to be consistent. Also `jwplayer-fandom` lives on borrowed time so I wouldn't care that much about naming ;)

src/jwplayer/utils/updateRVParam.ts
andrzejkupczyk: Personally, I have more confidence in the results given by the URL API rather than manipulating the URL in a customized way. It might be a good idea to verify this in the future, but it doesn't have to be done with this particular PR.

latata: There is no future for jwplayer-fandom ;)

andrzejkupczyk: This code is currently in production, and the repository is not being archived. It is important to note that the same assurance was previously given about Connatix.

dzawada (approved):

src/jwplayer/utils/updateRVParam.ts
```diff
@@ -0,0 +1,29 @@
+export function updateRVParam(adTagUrl: string, rv: number) {
```
karimMourra, line 1: I believe you could use [URL SearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) to simplify the code. Though not sure what the performance cost would be since you would be:
- transforming the URL string to a URL object
- using the URL object's searchParams API to read and write `rv` 
- converting back to string

PR #241, opened 2024-07-10
COTECH-1256 PoC for Newrelic integration

nandy-andy:
The easiest way to set custom attributes which we can use in NewRelic. We can test it live on the three test wikis and decide if the code should live somewhere else at the end.

### DIFF
--- a/src/jwplayer/utils/useScript.ts
+++ b/src/jwplayer/utils/useScript.ts
@@ -4,19 +4,19 @@ interface ScriptOptions {
 	id?: string;
 	async?: boolean;
 	className?: string;
-	onLoad: () => void;
+	onLoad?: () => void;
 }
 
 export default function useScript(
 	url: string,
 	parentNode: HTMLElement,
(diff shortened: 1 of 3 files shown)


andrzejkupczyk (approved):

src/jwplayer/utils/useScript.ts
```diff
@@ -4,19 +4,19 @@ interface ScriptOptions {
...
 	parentNode: HTMLElement,
-	beforeLoad: () => void,
+	beforeLoad: () => void | null,
 	scriptOptions: ScriptOptions,
 ) {
 	const [scriptInjected, setScriptInjected] = useState(false);
 
 	useEffect(() => {
-		beforeLoad();
+		typeof beforeLoad === 'function' ? beforeLoad() : null;
```
andrzejkupczyk, line 19: Personally, I’d prefer an `if` statement here instead of the ternary operator 😉

